PR #5, opened 2021-03-09
Use COPY instruction

szepeviktor:
No shell commands in Dockerfile


jaydrogers:
Just testing for now... It should be removed soon 😁👍

Thanks though! Lots more coming soon!!

jaydrogers:
Would love to have your feedback once I have it more populated. Thanks!

PR #43, opened 2022-01-19
Give granular controls on Laravel automations

jaydrogers:
# Problem
* `RUN_LARAVEL_AUTOMATIONS` is defaulted to `true`, which is not ideal for large environments
* Changing `RUN_LARAVEL_AUTOMATIONS` to `false` will disable other important features for containers

# Proposed solution
- [x] Remove the `RUN_LARAVEL_AUTOMATIONS` variable
- [x] Create a new variable called `AUTORUN_LARAVEL_MIGRATION` and default it to `false`
- [x] Create a new variable called `AUTORUN_LARAVEL_STORAGE_LINK` and default it to `true`
- [x] Update README with new variables
- [ ] Publish important note to users in new release (expected, **v1.1.0**)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ You must have these installed on your system.
 1. All templates are stored in the `/templates` folder
 1. I have a Git "Pre-Commit" hook that runs `build.sh`
 1. `build.sh` copies the templates and applies the templates with [yasha](https://github.com/kblomqvist/yasha)
-1. All generated files are then stored in the `/php` folder
+1. All generated files are then stored in the `/generated-dockerfiles` folder
 1. Github Actions will read the generated files and build images from the generated files
 
 # Running things locally
--- a/README.md
+++ b/README.md
@@ -114,17 +114,19 @@ We have a ton of helpful scripts and security settings configured for managing L
 ### Automated tasks executed on every container start up
 We automatically look at your Laravel `.env` file and determine if theses tasks should be run. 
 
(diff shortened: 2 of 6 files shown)


jaydrogers:
@danpastori: Can you review above and let me know if you like this? I will specifically be modifying these lines:

https://github.com/serversideup/docker-php/blob/e1a4a8e99fe9a2e82ac5bca7131d0356425540e0/templates/fpm/etc/cont-init.d/50-laravel-automations#L13-L39

danpastori:
@jaydrogers I'd structure it a little bit differently. The goal would be to make it as clear as possible to the user what is happening and allow you to scale to more fine grained controls as they come up. I've pseudocoded it below since I'm not a bash wizard:

```
# Scope it down to if you even want to look at the fine grain controls.
IF( APP_ENV_SETTING != "local" && !-z APP_ENV_SETTING && CI_ENV != true )
    # We want to operate in fine grain control mode and listen to the preferences from the user.

    IF( AUTORUN_LARAVEL_MIGRATION )
        # Run laravel migrations
    FI

    IF( AUTORUN_LARAVEL_STORAGE_LINK )
        # Bind storage link
    FI

    IF( AUTORUN_LARAVEL_CACHE_CLEAR )
        # Clear Cache
    FI

    # Future enhancements here
ELSE
    # Skip it all
FI
```

jaydrogers:
Hey @danpastori, could we potentially jump on the #live-events video chat to talk more? I want to make sure we're clear on somethings: https://serversideup.net/discord

Issue #47, opened 2022-01-26
Laravel automations not working when not using .env file


dakira:
In certain cloud deployment scenarios, all of the environment is set on the hosting site (e.g. Azure App Service) so that the .env is empty / not needed. The following unfortunately relies on .env existing and being used.

https://github.com/serversideup/docker-php/blob/4233993f7367c3fd39789a5826917bdfd12b4436/templates/fpm/etc/cont-init.d/50-laravel-automations#L11

Optimally the correct precedence should be used (i.e. what phpdotenv also uses). Maybe check if environment variable exists and use it if it does, if not check if .env exists and source it, if it does.

jaydrogers:
I think the true discussion is down to this line: https://github.com/serversideup/docker-php/blob/4233993f7367c3fd39789a5826917bdfd12b4436/templates/fpm/etc/cont-init.d/50-laravel-automations#L14

This line was added when things were more automated.

### Proposed change
* Maybe I just remove condition all together? I don't see a need for it now since we can explicitly enable migrations now with `AUTORUN_LARAVEL_MIGRATION`. Meaning if you have `AUTORUN_LARAVEL_MIGRATION` set to true, it should attempt to run the migrations without checking what APP_ENV you are in.

### Opening up for comment
Does this sound good @dakira? @danpastori: What are your thoughts?

shinsenter:
@jaydrogers 
According to the information [from the official Laravel documentation](https://laravel.com/docs/8.x/configuration#additional-environment-files), Laravel determines if either the APP_ENV environment variable has been externally provided.

> Before loading your application's environment variables, Laravel determines if either the APP_ENV environment variable has been externally provided or if the --env CLI argument has been specified. If so, Laravel will attempt to load an .env.[APP_ENV] file if it exists. If it does not exist, the default .env file will be loaded.

I think it's better for letting user change the value of `APP_ENV` either from `docker-compose.yml` file or via the docker command. The default value should be set as `APP_ENV_SETTING=${APP_ENV:-production}`.

jaydrogers:
For sure, I totally agree with this.

My proposal is I think I can eliminate this line all together: https://github.com/serversideup/docker-php/blob/4233993f7367c3fd39789a5826917bdfd12b4436/templates/fpm/etc/cont-init.d/50-laravel-automations#L14

I don't think we need to check for any of this anymore. Thoughts?

shinsenter:
@jaydrogers 
Yes, I think you can remove that condition.

[...]

PR #49, opened 2022-01-26
Simplified automation logic

jaydrogers:
# Problem
**Related bug:** #47 

### Summary
- The logic is too complicated and prevents automations from running
- This mainly is clean up from the original logic when we first had these automations in there

# Proposed solution
- [x] Simplify logic on running the automations
- [x] Update README and ensure all variables are up to date or removed

### DIFF
--- a/templates/fpm/etc/cont-init.d/50-laravel-automations
+++ b/templates/fpm/etc/cont-init.d/50-laravel-automations
@@ -5,19 +5,12 @@ set -e
 
 # Check to see if an Artisan file exists and assume means Laravel is configured.
 if [ -f $WEBUSER_HOME/artisan ]; then
-    ############################################################################
-    # Environment detection
-    ############################################################################
-    APP_ENV_SETTING=$(cat $WEBUSER_HOME/.env | grep APP_ENV | cut -f2 -d"=")
-
-    # If Laravel is not running in development or CI, then automate some production tasks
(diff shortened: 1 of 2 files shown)


jaydrogers:
@danpastori: Let me know your thoughts above and if this looks good.

templates/fpm/etc/cont-init.d/50-laravel-automations
```diff
@@ -3,21 +3,14 @@
...
-            echo "🏃‍♂️ Checking for Laravel automations..."
+        echo "🏃‍♂️ Checking for Laravel automations..."
 
         ############################################################################
         # Automated database migrations
         ############################################################################
-        if [ "grep -q DB_DATABASE $WEBUSER_HOME/.env" ] && [ ${AUTORUN_LARAVEL_MIGRATION:="false"} == "true" ]; then
+        if [ ${AUTORUN_LARAVEL_MIGRATION:="false"} == "true" ]; then
             echo "🚀 Running migrations..."
             su - webuser -c "php $WEBUSER_HOME/artisan migrate --force"
```
shinsenter, line 15: Since some migration operations are destructive, which means they may cause us to lose data.
Honestly, I think `--force` flag is not a good idea for production.

shinsenter:
@jaydrogers LGTM!
Note: The migration script is great for small apps, but for large apps we recommend running all migrations manually.

jaydrogers:
Cleaned up the documentation as well.

@danpastori: Please take a review and let me know if this looks good to you and I will merge 🚀

danpastori (approved):
This all looks good to me! Placing the variables in the docker compose files is a great idea.

PR #51, opened 2022-01-26
Fix issues with Livewire

jaydrogers:
# Problem
- NGINX is trying to serve a static assets for anything `*.js`

# Proposed solution
- [x] Add `try_files $uri /index.php?$query_string;` to allow JS files to be passed server side

### DIFF
--- a/generated-dockerfiles/7.4/fpm-nginx/etc/nginx/server-opts.d/performance.conf
+++ b/generated-dockerfiles/7.4/fpm-nginx/etc/nginx/server-opts.d/performance.conf
@@ -14,6 +14,7 @@ location = /robots.txt {
 location ~* \.(?:css(\.map)?|js(\.map)?|jpe?g|png|gif|ico|cur|heic|webp|tiff?|mp3|m4a|aac|ogg|midi?|wav|mp4|mov|webm|mpe?g|avi|ogv|flv|wmv)$ {
     expires    7d;
     access_log off;
+    try_files $uri /index.php?$query_string; 
 }
 
 # svg, fonts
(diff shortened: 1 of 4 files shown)


shinsenter:
@jaydrogers 
Personally, I think the following two options should go together as a set for static files.
```nginx
log_not_found off; access_log off;
```

generated-dockerfiles/7.4/fpm-nginx/etc/nginx/server-opts.d/performance.conf
```diff
@@ -14,6 +14,8 @@ location = /robots.txt {
 location ~* \.(?:css(\.map)?|js(\.map)?|jpe?g|png|gif|ico|cur|heic|webp|tiff?|mp3|m4a|aac|ogg|midi?|wav|mp4|mov|webm|mpe?g|avi|ogv|flv|wmv)$ {
     expires    7d;
     access_log off;
+    # Pass to PHP to help support things like Laravel Livewire
```
dakira, line 17: I propose changing the comment to

`# Pass to PHP to ensure PHP apps can handle routes that end in these filetypes`

as this is neither Livewire, nor Laravel specific. The same would happen if you server a route ending in .css with e.g. Yii.

shinsenter: Removing those comments is better, thought.

jaydrogers: I love having the comment in there because it reminds me on why it's there 😆

I also added this:
```
access_log off;
log_not_found off;
```

Thanks guys!

shinsenter:
@jaydrogers 
You haven't fixed it completely in generated-dockerfiles before merging...

jaydrogers:
LOL, dammit 🤦‍♂️

My pre-commit script didn't run for some reason.

Just fixed that. Thanks for pointing that out!

PR #52, opened 2022-03-11
Added environment variable to extend opinionated `open_basedir` measure

herpaderpaldent:
This PR adds the ability to extend the `open_basedir` setting with an environment variable

## Usecase
I develop on a project that splits in multiple composer packages with a common root/core package based on laravel. In order to develop/commit on the packages i have them stored in another dir and symlink the packages into the vendor folder with composers repositories setting. 

This is only done during development. Hence during development i run into `open_basedir` restrictions errors because other then `WEBUSER_HOME` no other paths are allowed. 
image

This environment var can be used in the `docker-compose.dev.yml` f.e. 

```yml
services:
  environment:
    PHP_BASEDIR_EXTENSION='/var/www/packages'
```

## Testcase

✅ Throws `base_dir` restriction error if no PHP_BASEDIR_EXTENSION variable is set
✅ No restriction error is been thrown if variable is set (to /var/www/packages)


jaydrogers:
So unfortunately I made a huge change to the structure of the repository, which complicated the merge.

Good news is, this feature is available on the [dev](https://github.com/serversideup/docker-php/tree/dev) branch (or the beta images):  https://hub.docker.com/r/serversideup/php/tags?name=beta-&page=1&ordering=-name

Let me know if this works to expectation. You should be able to find the documentation in the README on the **dev** branch. 👍

I appreciate your help & efforts!

PR #54, opened 2022-04-06
update examples to declare commands in list manner

herpaderpaldent:
# Problem
when running a container with a command in _shell_ form the passed environmental variables are not  available because the command is executed by the shell and environment variable expansion will be done by the  shell, not docker. 

[Docker Documentation](https://docs.docker.com/engine/reference/builder/#cmd)

If you declare environmental variables f.e. the `REDIS_HOST` or `DB_HOST` variable in your docker-compose and then run your queue worker, the queue worker won't be using these set variables but instead the variables present to the shell.

# Solution
By writing the command in the _list_ or _array form_, the passed environmental variables will be present at runtime of the command. Furthermore, docker declares the _array form_ as its preferred format of  `CMD`. This change hence follows best practices from docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,9 @@ services:
 
   task:
     image: my/laravel-app
-      # Switch to "webuser" before running `php artisan`
-    command: "su - webuser -c \"php artisan schedule:work\""
+    # Switch to "webuser" before running `php artisan`
+    # Declare command in list manner for environment variable expansion
+    command: ["su", "webuser", "-c", "php artisan schedule:work"]
     environment:
       PHP_POOL_NAME: "my-app_task"
 ```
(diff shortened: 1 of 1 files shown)


jaydrogers:
Holy crap! What a pain to trace this down. Thanks for pointing this out. I would have never guessed that 🙈

Let me do some testing on this before I merge.  Thanks for submitting this.

shinsenter:
I think the main reason is the hyphen between `su - webuser` and `su webuser`.

herpaderpaldent:
> I think the main reason is the hyphen between `su - webuser` and `su webuser`.

No, it's all about the presence of environmental variables. If you execute it in the shell variant it is as executing a shell directly. The environmental variable expansion is done from shell not docker (read the linked documentation regarding this mechanism). 

@jaydrogers had it described as:
`So wrapping it in a list preserves the Environment variables?`
https://discordapp.com/channels/910287105714954251/910299240897605632/960915300499472474 
and that's simply put it

jaydrogers:
After having more time to research this, I like this PR as-is. Thanks @herpaderpaldent for your contributions and your efforts to research the issue, provide a solution, and link to a factual document.

Thanks a ton!

PR #56, opened 2022-06-22
add php intl package to support laravel email validation rules

skylerkatz:
Hello!

Thanks for this great repository.

While using Laravel's [email validation rules](https://laravel.com/docs/9.x/validation#rule-email) I end up with an exception because the underlying package that they use requires the intl package to be installed.

I created a custom Docker image, but I figured I would submit this as a PR since these images are tuned for Laravel.


jaydrogers:
Awesome @skylerkatz! Thanks for your PR!

Issue #105, opened 2022-12-13
Add another variable that allows to toggle session.cookie_secure


zigazajc007:
### Affected Docker Images

All of them

### Current Behavior

session.cookie_secure = 1 is currently hard-codded and can't be changed with variables.

### Expected Behavior

Another variable that can change session.cookie_secure

### Steps To Reproduce

Every cookie will include secure tag, so cookies won't work on insecure / http websites.

### Host Operating System

Fedora 37

### Docker Version

```markdown
Version:           20.10.21
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        baeda1f
 Built:             Tue Oct 25 18:02:36 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
```

jaydrogers:
This is resolved in the v3 of the Docker images (coming soon).

It was resolved in this commit: https://github.com/serversideup/docker-php/commit/c8d848a72d007f86168360d322f2822f98669869

Status of the v3 is located here: https://github.com/serversideup/docker-php/pull/207

PR #107, opened 2022-12-30
Add PHP_SESSION_COOKIE_SECURE

jaydrogers:
### Related issues
- #105


jaydrogers:
@zigazajc007: I have this as a work-in-progress, but I am running into an error.

## Problem
- Setting `true` for `session.cookie_secure` via a Docker environment variable doesnt work

## Workaround, but then fails
- If I set it to `1`, it works
- If I change it to `0`, it fails

## Steps to reproduce
1. Download my `secure-cookie` branch from my test app: https://github.com/jaydrogers/docker-php-test-app/tree/secure-cookie
2. Change the environment variable from `0` to `1`

## Results

#### When `PHP_SESSION_COOKIE_SECURE = 1`
Everything works as expected ✅

**Docker Output:**
Screenshot 2022-12-29 at 21 16 32

**PHP Info:**
Screenshot 2022-12-29 at 21 16 48

#### When `PHP_SESSION_COOKIE_SECURE = 0`
Docker fails to start ❌

**Docker Output:**
```
ERROR: [/etc/php/8.1/fpm/pool.d/y-override-php-defaults.conf:227] error while parsing 'php_flag[session.cookie_secure]' : invalid boolean value
```
Screenshot 2022-12-29 at 21 17 23

## Next steps
- Do you have any thoughts on how we can get this to work? It doesn't seem to like the variables when I pass them through Docker environment variables

zigazajc007:
Hello,

I'm not sure if Docker environment variables even support booleans. It will probably just provide a variable in string format.

Did you tried to use On and Off instead?

zigazajc007:
I would personally provide image with php.ini that has most of security features turned on by default. So the ones that doesn't know how to customize php.ini then it shouldn't impact security if they use default php.ini that you provide.

But for those who knows how to edit php.ini and are willing to provide their own then I wouldn't force settings in some hidden .ini files that overwrite the default php.ini

Advanced developers would think that your image is broken if they try to turn off secure cookies and it won't work.

So I will just make sure that the default php settings are secure, but still leave developers to make any change with ease.

jaydrogers:
Another option is to leave it as-is, then someone can add a `zzz` config file that will be loaded after mine: https://github.com/serversideup/docker-php/tree/main/src/fpm/etc/php/fpm/pool.d

zigazajc007:
What if you check if user did set SECURE_COOKIES in env variable to false, Off or 0 and if so, then you can just append new line in your .conf file with echo?

zigazajc007:
It seems to be only working if I delete this line from the file:
```bash
RUN sed -i '/session.cookie_secure/d' /etc/php/current_version/fpm/pool.d/y-override-php-defaults.conf
```

jaydrogers:
This PR is being closed in favor for the fix being applied in the v3 of the Docker images (coming soon).

It was resolved in this commit, where I moved the config out of PHP-FPM and set it within the PHP.ini: https://github.com/serversideup/docker-php/commit/c8d848a72d007f86168360d322f2822f98669869

Status of the v3 is located here: https://github.com/serversideup/docker-php/pull/207

PR #172, opened 2023-04-05
Laravel Autorun: Fixing folder permission on the fly

ramaID:
Can we have this? I'm facing this issue.

Permission denied on symlink

### DIFF
--- a/src/fpm/etc/s6-overlay/scripts/laravel-automations
+++ b/src/fpm/etc/s6-overlay/scripts/laravel-automations
@@ -8,6 +8,14 @@ if [ -f $WEBUSER_HOME/artisan ] && [ ${AUTORUN_ENABLED:="true"} == "true" ]; the
         echo "🏃‍♂️ Checking for Laravel automations..."
 
         ############################################################################
+        # Automated fixing folder permission
+        ############################################################################
+        if [ ${AUTORUN_LARAVEL_PERMISSION:="true"} == "true" ]; then
+            echo "🔑 Fixing folder permission..."
+            s6-setuidgid root chown -R webuser:webgroup /var/www/html/storage /var/www/html/bootstrap/cache
+        fi
+
+        ############################################################################
         # Automated database migrations
         ############################################################################
         if [ ${AUTORUN_LARAVEL_MIGRATION:="false"} == "true" ]; then


jaydrogers:
I like where this is going and I appreciate that you used the S6 Overlay methods to do this too. I've had this happen to me a few times before too.

Are you able to give me steps to replicate the error you have above? I'd like to do some testing on my end too.

Also, do you think it we should have the directories controlled by an environment variable but default it? I wonder if we need to check if the directories exist too (to prevent warnings or errors)

XaiMeS:
I got this error on a clean installation of Laravel via composer on Ubuntu WSL2

jaydrogers:
Thanks, this probably all stems from this command:

```
composer create-project laravel/laravel example-app
```

When you run that, it defaults it as `root` (not `webuser`). I also cannot use the `USER webuser` directive in the Dockerfile because `root` is required to spawn services.

The PR proposed is one solution, but there are also deeper levels to solve this.

I'm also afraid the container will still run into an issue even after executing because it will:

1. Start the container
2. Run the permission fix
3. Run composer as root
4. Install everything as root

As you can see, the fix is running at step #2, where it needs to run after step #4.

S6 Overlay provided things like this in their v2, but you can see they have made changes: https://github.com/just-containers/s6-overlay#fixing-ownership-and-permissions

This permissions issue is something I've been aware of for sometime, but I just haven't thought of the best solution yet.

krzysztof97:
### Simple solution
Set env variables `PUID` and `PGID` to values corresoponding to your user on host machine in `docker-compose.yml`.
The first unix user created will almost always have a value of `1000`.

Just type `cat /etc/passwd` in terminal and search for values near your user name, e.g. `krzysztof:x:1000:1000:,,,:/home/krzysztof:/bin/bash`

Then set these values in environment section of your `docker-compose.yml` file:

```
environment:
    PUID: 1000
    PGID: 1000
```

### Explanation
You've linked _project directory_ to `/var/www/html` in container. By default webserver run as user with UID and GID set to 9999. Container has automation script which links storage using artisan command, but cannot create symlink on your host machine when script is running as user with UID not corresponding to your UID. Unfortunately, it is impossible to create a symlink only in the container when the directory is on the host.

jaydrogers:
> Set env variables PUID and PGID to values corresoponding to your user on host machine in docker-compose.yml

Even with setting these variables, a Docker run command will default to `root`. This is because the container needs root to bring up NGINX + FPM.

Test it out yourself with:
```
docker run --rm serversideup/php:8.2-fpm-nginx /bin/bash -c "composer create-project laravel/laravel example-app; ls -al /var/www/html"
```

You can see the `ls -al` of `/var/www/html` shows the example app is created with `root` permissions:
```
drwxr-xr-x  1 webuser webgroup 4096 Apr 18 14:17 .
drwxr-xr-x  1 root    root     4096 Apr 18 08:26 ..
drwxr-xr-x 12 root    root     4096 Apr 18 14:17 example-app
```

jaydrogers:
Just adding a note it was on my list to investigate this today 😃

Tracking it here: https://github.com/serversideup/docker-php/issues/179

Issue #179, opened 2023-05-24
Failed to open stream: Permission denied


jaydrogers:
### Affected Docker Images

All

### Current Behavior

There are many reports where this error randomly appears, usually on fresh deployments.

```
Failed to open stream: Permission denied
```

#### Examples
https://github.com/serversideup/docker-php/pull/172#issue-1654792968

**Bugflow**
<img width="1723" alt="image" src="https://github.com/serversideup/docker-php/assets/3174134/8c1bb3fe-6a75-4037-be87-5641768f064e">

### Expected Behavior

There is a discussion where we may be able to adjust S6 Overlay and have the commands run correctly (instead of root)

https://github.com/serversideup/docker-php/discussions/71#discussioncomment-5841649

### Steps To Reproduce

TBD

### Host Operating System

Linux

### Docker Version

```markdown
N/A
```

jaydrogers:
Adding my error message in here too for future reference 


<img width="1949" alt="image" src="https://github.com/serversideup/docker-php/assets/3174134/9ecd7ddb-935b-4958-b13e-2624e1e4d309">

danilopolani:
Wanted to add a few things on this issue.  
I copied all the project files with the proper user ownership like this:

```Dockerfile
COPY --chown=$PUID:$PGID ./composer.json ./composer.lock ./
RUN composer install --no-cache --no-dev --no-scripts --no-autoloader --ansi --no-interaction

COPY --chown=$PUID:$PGID . .
RUN composer dump-autoload -o
``` 

Now the issue is that it can't create cache files:

> Unable to create lockable file: /var/www/html/storage/framework/cache/data/  
  fa/f2/faf26108316c685830630fa2070ff0a7b212a06a. Please ensure you have perm  
  ission to create files in this location.

I've tried giving to the whole `storage` folder all the permissions and ownership to both `$PUID:$PGID` or `www-data:www-data`, but no luck so far. Also with the `beta` I got no luck. What's the user running the webserver?

jaydrogers:
Thanks for chiming in. The user running PHP is `webuser`, so you might need to do a `su` before running your composer commands.

Also -- I am not sure why you would need two copy commands in there. Also, the default web root is `/var/www/html`. If you're running your build within the Docker build process, then it would look something like this:

```Dockerfile
COPY --chown=$PUID:$PGID ./composer.json ./composer.lock /var/www/html
RUN su webuser && \
[...]

PR #180, opened 2023-05-24
Set `webuser` to as the default user for "run" commands

jaydrogers:
### Problem
- If you run `docker compose run php composer install`, it will pass the command and execute it as `root`
- This causes other issues to rise when `webuser` cannot access a log or the vendor folder in other containers

### What this PR does
This follows the methods detailed by @tsterker(https://github.com/serversideup/docker-php/discussions/71#discussioncomment-5841649)

```Dockerfile
ENV S6_CMD_ARG0='s6-setuidgid webuser'
```

### Possible breaking issues
- [ ] We need to ensure this change does not cause problems with our current suggested setup (which should be updated to remove the `su` command) :

```yaml
version: '3'
services:
  php:
    image: my/laravel-app
    environment:
      PHP_POOL_NAME: "my-app_php"

  queue:
    image: my/laravel-app
    # Switch to "webuser" before running `php artisan`
    # Declare command in list manner for environment variable expansion
    command: ["su", "webuser", "-c", "php artisan queue:work --tries=3"]
    environment:
      PHP_POOL_NAME: "my-app_queue"

```


jaydrogers:
# ⚠️UPDATE: This PR has been reverted
See this for more details: https://github.com/serversideup/docker-php/issues/179

PR #188, opened 2023-06-14
Enchance Laravel Automations

moisish:
Update the Laravel Automations for

- Route Caching
- Config Caching
- Event Discovery
- Optimizing/Cache Views

### DIFF
--- a/docs/content/docs/4.laravel/1.laravel-autorun-script.md
+++ b/docs/content/docs/4.laravel/1.laravel-autorun-script.md
@@ -31,3 +31,48 @@ php artisan migrate --force --isolated
 
 **Automatic migrations are DISABLED by default.** To enable, set an environment variable of `AUTORUN_LARAVEL_MIGRATION=true` on your container. This requires Laravel v9.38.0 or higher.
 
+::code-panel
+---
+label: Event Discovery
+---
+```bash
+php artisan event:cache
+```
+::
+
+**Event Discovery is ENABLED by default.** You can disable this behavior by setting `AUTORUN_LARAVEL_EVENT_CACHE=false`.
+
+
+::code-panel
+---
+label: Optimizing Views
+---
+```bash
+php artisan view:cache
+```
+::
+
+**Optimizing Views is ENABLED by default.** You can disable this behavior by setting `AUTORUN_LARAVEL_VIEW_CACHE=false`.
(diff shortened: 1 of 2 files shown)


alexjustesen:
2c on this PR, it would probably be better to run `artisan view:clear` to clear any cached views and then `artisan optimize` which does the route and cache config for you.

moisish:
Why do you want to clear the view cache? Also running them individual is more flexible since you can disable one

gtjamesa:
I've just discovered this repo, and after reading through the docs, I thought this would be a great addition. Luckily, someone had already submitted a PR 😁 I think configuring each cache is the better option, particularly as I have a legacy project that cannot cache its routes.

dakira:
I don't see why this would be a good idea. There is no benefit in re-doing the caching every time the image starts. The code doesn't change unless the image content changes. A better place (IMHO) would be to do the caching in your build process, right?

gtjamesa:
That's likely true for event/routes, but config will be caching environment variables that are set during deployment, so they can be modified post-build.

moisish:
@dakira Personally I run event and view cache during the docker build.  The config cache cannot run during the build because the env variables are jot available.

All the options are available in the script for the maximum flexibility and can be individually disabled

jaydrogers:
Just adding a note on this I will be adapting this PR to the new v3.0 images.

I also have a Twitter thread of related notes and discussions here: https://twitter.com/jaydrogers/status/1721574061111701630

src/fpm/etc/s6-overlay/scripts/laravel-automations
```diff
@@ -7,6 +7,28 @@ set -e
...
 
+        ############################################################################
+        # Automated Laravel Cache
+        ############################################################################
+        if [ ${AUTORUN_LARAVEL_CONFIG_CACHE:="true"} == "true" ]; then
+            echo "🗄️ Running config cache..."
+            s6-setuidgid webuser php $WEBUSER_HOME/artisan config:cache
+        fi
+
+        if [ ${AUTORUN_LARAVEL_VIEW_CACHE:="true"} == "true" ]; then
```
a-h-abid, line 18: I think this should be `AUTORUN_LARAVEL_EVENT_CACHE` ???

PR #287, opened 2024-02-28
Added gosu to os dependencies

sergiorodenas:
Gosu is needed for doing later `command: ["gosu", "www-data", "/usr/local/bin/php", "artisan", "horizon"]` as the 
_su_ version `command: ["su", "www-data", "-c", "/usr/local/bin/php", "artisan", "horizon"]` won't work due the _www-data_ user having the login disabled internally.


AlejandroAkbal:
Yeah, without this, it will not work, until it gets merged, this is a workaround:

Create a dockerfile with this

```dockerfile
FROM serversideup/php:beta-${PHP_VERSION}-fpm-nginx

# TODO: Fix for gosu, remove when this is merged: https://github.com/serversideup/docker-php/pull/287
RUN docker-php-serversideup-dep-install-debian "gosu"
```

jaydrogers:
Instead of the complexities with changing users, what do you think about this approach instead? https://github.com/serversideup/docker-php/issues/179#issuecomment-2052513529

AlejandroAkbal:
This can now be closed

PR #294, opened 2024-03-13
Install opcache extension by default

josiasmontag:
A production ready and optimized PHP image should really have the `opcache` extensions installed by default.

The performance gains are [massive](https://arubacao.com/laravel-opcache-performance-benchmark/) (especially for Laravel) and it is just installing the extension.


jaydrogers:
> What! Quite weird that opcache is not installed by default, this should be prioritized

Gimme a break guys! I'll get it merged 😆

anotherglitchinthematrix:
:^) what the hell, last time I checked with `php -m` I think I saw opcache listed there and I'm pretty sure I've configured opcache for my needs.

anotherglitchinthematrix:
Yes it's already here

```sh
> docker run --rm serversideup/php:8.2-fpm php -m | grep "Zend OPcache"

[18-Mar-2024 08:42:37] NOTICE: fpm is running, pid 64
[18-Mar-2024 08:42:37] NOTICE: ready to handle connections
[18-Mar-2024 08:42:37] NOTICE: systemd monitor interval set to 10000ms
Zend OPcache
Zend OPcache
[18-Mar-2024 08:42:38] NOTICE: Terminating ...
[18-Mar-2024 08:42:38] NOTICE: exiting, bye-bye!

> docker run --rm serversideup/php:8.2-cli php -m | grep "Zend OPcache"

Zend OPcache
Zend OPcache
```

AlejandroAkbal:
Maybe it’s not on the beta?

josiasmontag:
Indeed it seems to be enabled in stable but not in the beta.

```
➜  ~ docker run --rm -it serversideup/php:8.2-fpm-nginx php -i | grep opcache.enable
opcache.enable => On => On
opcache.enable_cli => Off => Off
opcache.enable_file_override => Off => Off
➜  ~ docker run --rm -it serversideup/php:beta-8.3-fpm-nginx php -i | grep opcache.enable
➜  ~ 
```

krzysztof97:
At this moment you can just add `RUN install-php-extensions opcache` to your own Dockerfile.

jaydrogers:
I'm working on this now if anyone is available to give me quick feedback:

Opcache will be installed.

**👉 The question:**

- Should it be ENABLED by default?

I am just worried about PHP developers running into local issues regarding cache. It will be easy to disable it via an ENV setting (which i will document.

Thoughts? 🤔

seanvandermolen:
I'd prefer it enabled, but for your sanity, probably disabled.

sneycampos:
i used to have development and production images, each one with his default tools, like xdebug in dev, opcache in prod. Documenting the path to customize these settings etc.

jaydrogers:
Just shipped this to the `beta` images.

Everything is fully documented on the live site and I put a notification if it's enabled or disabled.

image

### How it works
- `PHP_OPCACHE_ENABLE` is set to `0` by default
- A notice will display on how to improve performance with a link to the docs if it's disabled

### Why I did it this way
- We don't have to maintain two image bases
- The user can still make these changes easily on a per environment basis with a multi-stage Dockerfile with proper targets or Docker Compose overrides

Issue #304, opened 2024-04-12
Upstream sent too big header while reading response header from upstream


jaydrogers:
### Affected Docker Images

beta-*-fpm-nginx

### Current Behavior

When using IntertiaJS, some pages experience a 502 error with this in the logs:

```
upstream sent too big header while reading response header from upstream
```

<img width="735" alt="image" src="https://github.com/serversideup/docker-php/assets/3174134/3167ab23-263d-4966-bca3-566113ce0a58">

### Expected Behavior

The page should not have 502 errors

### Steps To Reproduce

N/A

### Host Operating System

N/A

### Docker Version

```markdown
N/A
```

PR #305, opened 2024-04-12
Set default NGINX values for "fastcgi_buffers" and "fastcgi_buffer_size"

jaydrogers:
# Fixes
- #304 

# What this PR does
- Set `fastcgi_buffers` to `8 8k`
- Set `fastcgi_buffer_size` to `8k


bretto36:
Can these be changed into configurable options? Currently we have to replace them using a sed call in the Docker file

We use InertiaJS and it sends a bunch of extra headers which keep blocking nginx

@jaydrogers

PR #367, opened 2024-05-18
Improve os-release file processing

szepeviktor:
From https://www.freedesktop.org/software/systemd/man/latest/os-release.html#Examples


szepeviktor:
Thank you for merging.

jaydrogers:
You bet! Thanks for your continued contributions 🙌

PR #374, opened 2024-05-31
Add an option to disable/redirect fpm access logs

RadeJR:
Since the introduction of healthchecks in new images, my logs are very crowded with /healthcheck endpoint so I added an option to change access log file with environment variable. For my use case, I don't really need fpm access logs, for laravel they are just requests to /index.php route anyway, nginx access logs show more info, in that case we don't need duplicates.


jaydrogers:
Thanks for pointing this out! I ended up removing the access log all together unless it's in Debug Mode.

I silenced the Apache Access Log for the healthcheck as well 👍

Issue #381, opened 2024-06-18
Allow FastCGI parameters to be set via environment variable


jaydrogers:
Can these be changed into configurable options? Currently we have to replace them using a sed call in the Docker file

We use InertiaJS and it sends a bunch of extra headers which keep blocking nginx

@jaydrogers

_Originally posted by @bretto36 in https://github.com/serversideup/docker-php/issues/305#issuecomment-2148840930_

jaydrogers:
Hey @bretto36,

Could you elaborate more on what issues you're running into? We're running InertiaJS and we haven't had any issues with it.

bretto36:
On pages with lots of components the headers that get sent for preloading from laravel/inertia cause an error

We've had to increase the fastcgi size a few times now. Only happens on pages with LOTS of components though. 

We also have a large CSP in place which probably adds to the overall header size

bretto36:
@jaydrogers  This is what we do to fix currently and it's not clean. Because we bad a 502 bad gateway from nginx

#FIND fastcgi_buffers 8 8k; and change to fastcgi_buffers 8 128k;
RUN sed -i 's/fastcgi_buffers 8 8k;/fastcgi_buffers 8 128k;/' /etc/nginx/site-opts.d/http.conf.template

#FIND fastcgi_buffer_size 8k; and change to fastcgi_buffer_size 128k;
RUN sed -i 's/fastcgi_buffer_size 8k;/fastcgi_buffer_size 128k;/' /etc/nginx/site-opts.d/http.conf.template

This is the header that comes through on inertia requests and on pages where there might be 100 components it's causing the 502 bad gateway
Link:
[...]

PR #382, opened 2024-06-18
Allow fastcgi parameters to be set via environment variable

jaydrogers:
### Fixes
- https://github.com/serversideup/docker-php/issues/381


bretto36:
@jaydrogers Thanks so much! Really appreciate it!

bretto36 (approved):

PR #390, opened 2024-06-28
Add FrankenPHP support - Update Dockerfile

sneycampos:
- set `ENV SERVER_NAME=:8080` so frankenphp can run in 8080 port
- using frankenphp base `install-php-extensions` script to install php extensions (and system dependencies?)
- removed usage of `docker-php-serversideup-dep-install-debian` since all dependencies are installed by frankenphp `install-php-extensions`

### DIFF
--- a/src/variations/frankenphp/Dockerfile
+++ b/src/variations/frankenphp/Dockerfile
@@ -10,7 +10,6 @@ ARG BASE_IMAGE="dunglas/frankenphp:${FRANKEN_PHP_VERSION}-php${PHP_VERSION}-${BA
 FROM ${BASE_IMAGE}
 
 ARG DEPENDENCY_PACKAGES_ALPINE='shadow'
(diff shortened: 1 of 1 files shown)


sneycampos:
@jaydrogers Should we remove capabilities from Frankenphp since its [running in a non privileged port](https://frankenphp.dev/docs/docker/#running-with-no-capabilities)?

DarkGhostHunter:
It should be aware that FrankenPHP only works with PHP 8.1 (given what the docs say, haven't tested on older versions and even if it does it may have zero support)

sneycampos:
> It should be aware that FrankenPHP only works with PHP 8.1 (given what the docs say, haven't tested on older versions and even if it does it may have zero support)

FrankenPHP does not support PHP 8.1 and won't: [#637](https://github.com/dunglas/frankenphp/issues/637)

DarkGhostHunter:
Thanks for the correction. Should be 8.2.

PR #392, opened 2024-07-03
changed case of `as` to `AS` to stay inline with convention

chinmaypurav:
https://docs.docker.com/reference/dockerfile/#format

`However, convention is for them to be UPPERCASE to distinguish them from arguments more easily.`


jaydrogers:
Thanks for the suggestion!

20x-dz:
The different cases also trigger a warning when launching the containers, see https://docs.docker.com/reference/build-checks/from-as-casing/

Thank you for adjusting the styles.

PR #393, opened 2024-07-08
Adds note for using cron

DarkGhostHunter:
Adds some paragraphs to the docs regarding using `cron` as an alternative when there is no other alternative.

On Debian images, it must be installed manually (usually on the production step), while on Alpine images `crond` is included OOTB.

Sometimes there is no possibility to add a scheduler command, like on third-party applications or else.


docs/content/docs/4.laravel/2.laravel-task-scheduler.md
```diff
@@ -16,6 +16,94 @@ Running a Laravel task scheduler with Docker can be a little different from the
 ## More detail
 We need to run the [schedule:work](https://laravel.com/docs/11.x/scheduling#running-the-scheduler-locally) command from Laravel. Although the docs say "Running the scheduler locally", this is what we want in production. It will run the scheduler in the foreground and execute it every minute. You can configure your Laravel app for the exact time that a command should run through a [scheduled task](https://laravel.com/docs/11.x/scheduling#scheduling-artisan-commands).
 
+## Using cron
+Laraver versions prior to 8.x don't have the `schedule:work` command, as [noted by this PR](https://github.com/laravel/framework/pull/34618).
```
adrianmejias, line 20: Laraver -> Laravel

eaguirrek12com:
Thank you!

eaguirrek12com:
**Hi @DarkGhostHunter, I’m having trouble getting this setup to work.**

  

Here’s my docker-compose.yml setup:

```
services:
  php:
    build:
      context: .
      dockerfile: Dockerfile
    image: php_una
    environment:
      APACHE_DOCUMENT_ROOT: "/var/www/html"
      PHP_FPM_POOL_NAME: "php_una"
    ports:
      - "80:8080"
      - "443:8443"

  php_cron:
    image: php_una
    environment:
      APACHE_DOCUMENT_ROOT: "/var/www/html"
      PHP_FPM_POOL_NAME: "php_una"
    command: [ "/usr/sbin/cron", "-f", "-L", "15" ]
```

And my Dockerfile:

```
FROM serversideup/php:8.1-fpm-apache

USER root

RUN apt-get update && apt-get install -y \
    ffmpeg \
    cron \
    && rm -rf /var/lib/apt/lists/*

# Install PHP extensions
RUN install-php-extensions \
    curl \
    gd \
    mbstring \
    json \
    fileinfo \
    zip \
    openssl \
    exif

# Default crontab file running as the "www-data" user
RUN printf "* * * * * www-data php -q /var/www/html/periodic/cron.php >/proc/1/fd/1 2>/proc/1/fd/2\n" >> /etc/crontab

# Copy application files as www-data (33:33)
COPY --chown=www-data:www-data ./app /var/www/html

# Drop back to unprivileged user
USER www-data

COPY php.ini /usr/local/etc/php/conf.d/zzz-custom-php.ini
```

However, when running the php_cron service, I encounter the following issue in the logs:

```
ℹ️  [NOTICE]: Running custom command instead of web server configuration: '/usr/sbin/cron -f -L 15'
seteuid: Operation not permitted
```

It looks like the cron process is failing due to permission issues (seteuid: Operation not permitted). Could this be due to running as www-data? Do you have any recommendations on how to properly configure the cron service within this container setup?

DarkGhostHunter:
I believe the image must change the cron permissions to be able to run.

Can you try adding this before the `USER www-data` line?

```
chmod gu+s /usr/sbin/cron
chmod gu+rw /var/run
```

eaguirrek12com:
yes, thank you

> I believe the image must change the cron permissions to be able to run.
> 
> Can you try adding this before the `USER www-data` line?
> 
> ```
> chmod gu+s /usr/sbin/cron
> chmod gu+rw /var/run
> ```

eaguirrek12com:
@DarkGhostHunter 
Thanks for the suggestion! I initially thought that modifying cron’s permissions with:  

```bash
chmod gu+s /usr/sbin/cron
chmod gu+rw /var/run
```

would allow www-data to run cron, and while this does enable execution, it doesn’t solve the core issue.

  

After further troubleshooting, I found that cron is designed to run as root and requires root privileges to function properly. Instead, I used the following approach:
  

**1️⃣ Installed cron as root in the Dockerfile.**

  

**2️⃣ Created a crontab for the www-data user:**

```
# Switch back to www-data user
USER www-data
# Set up www-data user crontab
COPY --chown=www-data:www-data path/to/your/www-data-crontab /tmp/crontab
RUN crontab /tmp/crontab && rm -f /tmp/crontab
```

**3️⃣ Ensured the container runs as root in docker-compose.yml:**

```
php_cron:
  image: yourDockerFileImage
  user: root # Required for cron to function correctly
  command: [ "cron", "-f" ]
```

This way, cron runs with the necessary permissions, but scheduled jobs still execute as www-data, maintaining proper security and access control.

mbrodala:
IMO one should not use Cron but [Supercronic](https://github.com/aptible/supercronic).

Install it in your `Dockerfile` as described (e.g. to `/usr/local/bin/supercronic`) and add a `crontab` file (e.g. `/etc/crontab`).

Then use a `command` like this:

```sh
/usr/local/bin/supercronic \
    -overlapping \
    -prometheus-listen-address 127.0.0.1:9746 \
    /etc/crontab
```

… and a `healthcheck` like this:

```sh
curl \
    --silent \
    --show-error \
    --fail \
    http://127.0.0.1:9746/health \
|| exit 1
```

Not sure if it is worthwhile to have this image ship Supercronic by default. One may also consider building a separate image variant tailored for Cron.

PR #398, opened 2024-07-11
fix docs regarding `AUTORUN_LARAVEL_MIGRATION_ISOLATION`

florian-nolte:
The default for `AUTORUN_LARAVEL_MIGRATION_ISOLATION` seems to be `false`, see https://github.com/serversideup/docker-php/blob/9999c0967575f66d8dd55f37f025f98885efde9c/src/common/etc/entrypoint.d/50-laravel-automations.sh#L73


jaydrogers:
Thanks for catching that! I forgot I had to change that some time ago.

PR #283, opened 2024-02-08
Release 4.0: Add FrankenPHP & more Laravel optmizations

jaydrogers:
# 4.0 Release: Say hello to FrankenPHP 👋
<div align="center"></div>

# 🚀 What this PR does
This PR is our home base for testing our new 4.0 release. This adds the highly anticipated variation of FrankenPHP and further optimizes _serversideup/php_ to be highly optimized for deploying and maintaining Laravel applications.

# 🫵 WE NEED YOU: Help us test test this release
> [!IMPORTANT]  
> **Please keep reading the notes in this post before cowboy coding and throwing this into production 🤠**
> 1. This PR is in an "alpha" state and could change at anytime
> 2. The images are located on a special "development" repository `php-dev` (not `php`)
> 3. There are more changes than just adding FrankenPHP. Read the entire post for what's new

## [View Test Images on Dockerhub →](https://hub.docker.com/r/serversideup/php-dev/tags?name=283-)
Anything tagged with `283-` will reference this PR and should be tested. Notice we're testing on `php-dev` (not `php`).

```
serversideup/php-dev:283-*
```

# 🐛 Reporting Issues
- [Open an issue on GitHub](https://github.com/serversideup/docker-php/issues/new) that is simple and reproducible. 
- Make sure it follows our "[Guide On How To Get Help](https://serversideup.net/ask-for-help/)".
- Clearly indicate it relates to the new 4.0 testing

# 👨‍🔬 What to test
We really need the community's help on testing these images as we progress towards stable.

### 1️⃣ All variations: Laravel Automations
There were huge improvements made to the [50-laravel-automations.sh](https://github.com/serversideup/docker-php/blob/280-create-a-frankenphp-variation/src/common/etc/entrypoint.d/50-laravel-automations.sh) script.

1. Does the script still work after upgrading an existing FPM-NGINX image?
2. Is the script behaving well and giving you the results you anticipate?

### 2️⃣ FrankenPHP: Caddyfile structure
1. Do you like the structure of the [Caddyfile](https://github.com/serversideup/docker-php/tree/280-create-a-frankenphp-variation/src/variations/frankenphp/etc/frankenphp)?
2. Do you find the script flexible and easy to expand?
3. Do the [performance defaults](https://github.com/serversideup/docker-php/blob/a3d592c5e94e2076ddf2604fad43b468139b3b27/src/variations/frankenphp/etc/frankenphp/Caddyfile#L84-L106) make sense and follow best practices?
[...]

PR #400, opened 2024-07-11
Add creating missing SQLite database files

DarkGhostHunter:
This PR allows the container to start and create missing the SQLite databases set by the application.

It's part of the container `AUTORUN_LARAVEL` scripts. It runs before the migrations.

The way it works is through the `AUTORUN_LARAVEL_TOUCH_SQLITE` environment value, which by default is `true`:

- When `true`, `1` or `default`, it will pick the default connection.
- When used as `default, telescope, custom`, it will find the connection names, and create each database file.

This is done by checking if the driver is `sqlite` and the file doesn't exist. If it's not SQLite, it fails. If the database file already exists, doesn't do anything.

### DIFF
--- a/src/common/etc/entrypoint.d/50-laravel-automations.sh
+++ b/src/common/etc/entrypoint.d/50-laravel-automations.sh
@@ -73,7 +73,7 @@ touch_sqlite_database() {
 
 # Set default values for Laravel automations
 : "${AUTORUN_ENABLED:=false}"
-: "${AUTORUN_LARAVEL_TOUCH_SQLITE:=false}"
+: "${AUTORUN_LARAVEL_TOUCH_SQLITE:=true}"
 : "${AUTORUN_LARAVEL_MIGRATION_TIMEOUT:=30}"
 
 if [ "$DISABLE_DEFAULT_CONFIG" = "false" ]; then
--- a/src/common/etc/entrypoint.d/50-laravel-automations.sh
+++ b/src/common/etc/entrypoint.d/50-laravel-automations.sh
@@ -45,15 +45,15 @@ touch_sqlite_database() {
 
         foreach (\$connections as \$name) {
             if (\$config->get(\"database.connections.\$name.driver\") !== 'sqlite') {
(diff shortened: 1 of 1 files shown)


jaydrogers:
Thanks for the PR on this. I just ran into this issue myself.

Regarding your comment:

> Newer Laravel versions do create the SQLite database file when force migrating. Otherwise, it won't.

Do you know what version this started happening in?

DarkGhostHunter:
I believe 9.x or 8.x, I'll have to see the blame.

jaydrogers:
Thanks!

If it goes back that far with Laravel, should we even merge this?

I was able to resolve my issue by having proper permissions on my Docker Volumes with Laravel 11. Once I had the correct permissions, the auomations worked perfectly and created my database.

DarkGhostHunter:
> Thanks!
> 
> 
> 
> If it goes back that far with Laravel, should we even merge this?
> 
> 
> 
> I was able to resolve my issue by having proper permissions on my Docker Volumes with Laravel 11. Once I had the correct permissions, the auomations worked perfectly and created my database.

9.31 https://github.com/laravel/framework/pull/44153

It doesn't go that far IMHO. In any case, I think older Laravel apps would benefit from this on staging and development, especially if you are on the path to upgrade.

jaydrogers:
Thanks for your efforts on this PR!

As I mentioned in the other post, I spent a ton of time refactoring the Laravel Automations script:

### What's new in my refactor
I decided to NOT include a "touch" for creating SQLite database file.

### My reason
Laravel will create this for you automatically with `php artisan migrate`. I did not want to step on Laravel's toes with this process.

Instead, you can see that I added helpful logic to help debug if a SQLite file does not exist when we test the DB connection: https://github.com/serversideup/docker-php/blob/280-create-a-frankenphp-variation/src/common/etc/entrypoint.d/lib/laravel/test-db-connection.php

Let me know your thoughts on my approach. I greatly appreciate your intentions on this PR and would gladly review any other improvements that you suggest 👍

PR #404, opened 2024-07-17
Add opt-int database seeding support

DarkGhostHunter:
This adds the ability to seed the database.

Laravel can seed the database without confirmation when the application is not on production. By using this ability to _confirm_, the seeding can be enabled by setting `AUTORUN_LARAVEL_SEED` to either `true` or the class name of the Database Seeder.

This step runs after the migration step.


chris0203:
This will not work in production. it got an error when APP_ENV=production

```shell
🚀 Running seeders...

                            APPLICATION IN PRODUCTION.                           

WARN  Command cancelled.
``` 

should change it to 
```php
php ${APP_BASE_DIR}/artisan db:seed --force
```

DarkGhostHunter:
> This will not work in production. it got an error when APP_ENV=production
> 
> 
> 
> ```shell
> 
> 🚀 Running seeders...
> 
> 
> 
>                             APPLICATION IN PRODUCTION.                           
> 
> 
> 
> WARN  Command cancelled.
> 
> ``` 
> 
> 
> 
> should change it to 
> 
> ```php
> 
> php ${APP_BASE_DIR}/artisan db:seed --force
> 
> ``` 

Shouldn't work on production. That's a feature. If you need seeding on production you have to do it manually.

harrisonratcliffe:
Would really like to see this implemented @jaydrogers

harrisonratcliffe:
> > This will not work in production. it got an error when APP_ENV=production
> > ```shell
> > 🚀 Running seeders...
> > 
> > 
> > 
> >                             APPLICATION IN PRODUCTION.                           
> > 
> > 
> > 
> > WARN  Command cancelled.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should change it to
> > ```
> > php ${APP_BASE_DIR}/artisan db:seed --force
> > ```
> 
> Shouldn't work on production. That's a feature. If you need seeding on production you have to do it manually.

Can add --force to get round that.

jaydrogers:
I see the use case for this now and I really like this. Let me evaluate this for v3.6 and I will keep you all posted 👍

jaydrogers:
Thanks for your contribution @DarkGhostHunter! This will be made available in v3.6: https://github.com/serversideup/docker-php/pull/523

aagjalpankaj:
@jaydrogers Just for my understanding, why seeding should be part of `50-laravel-automations.sh`?

In my opinion,
- Seeding can be a one-time activity for production and it's a very rare scenario. If someone want it, they should create a migration and call the seeder from that migration. I think that's the proper way for production.
- `50-laravel-automations.sh` will be triggerred from the entrypoint and will call the seeder everytime. That can break the deployment or may create unwanted records on production.

Though `AUTORUN_LARAVEL_SEED` is default false, I think seeding should not be part of the automation script. We should not make the image too extensive!

jaydrogers:
I think @aagjalpankaj is right on this.

I just spent a ton of time refactoring the Laravel Automations script and actually **removed `db:seed` support**.

### My reason
I agree with @aagjalpankaj that this command would be a one-time command that someone would likely run in the terminal.

### My different approach
Instead of using `db:seed`, I added support for "migration modes":

| Migration Mode | Description|
|---------------------|---------|
| `default` (our default behavior) | Runs `php artisan migrate` - standard forward migrations|
| `fresh` | Runs `php artisan migrate:fresh` - drops all tables and re-runs migrations|
| `refresh` | Runs `php artisan migrate:refresh` - rolls back and re-runs migrations|

This will help people who need to prepare testing environments.

**Still need to trigger your seeder in production?**
That's totally fine! We will just do it through the `--seed` flag on `php artisan migrate`: https://laravel.com/docs/12.x/migrations#roll-back-migrate-using-a-single-command

You can add the `--seed` flag by setting `AUTORUN_LARAVEL_MIGRATION_SEED` to `true`, which will indicate if the seed task should be re-run.

### Available to test now
These changes will be made available in v4.0, but you can test these changes now on this PR:

@DarkGhostHunter: Let me know if this approach works for your use case as well!

DarkGhostHunter:
Lol it works better:

- If I want to seed, it's because my database is already pristine.
- Otherwise, I use `fresh` or `refresh` and seed all over again.

Personally I believe that seeding should be always done with a pristine database. If the database exist, then you may be hitting production.

PR #407, opened 2024-07-30
Name service_docker-build-and-publish workflow

szepeviktor:
Just a name.


jaydrogers:
Thanks! Not sure if it makes a difference since I only call this file from other files, but it's nice to have it labeled 👍

szepeviktor:
The filename was ugly in the list:

image

jaydrogers:
Ahh, makes sense. Thanks!!

szepeviktor:
Glad to contribute.

PR #413, opened 2024-08-14
Fix #373: Run custom command in fpm images without running web server

### LINKED ISSUE #373: Nginx boots up and starts accepting connections when running a non default command.
### Affected Docker Images

```{
  "org.opencontainers.image.authors": "Jay Rogers (@jaydrogers)",
  "org.opencontainers.image.description": "Supercharge your PHP experience. Based off the official PHP images, serversideup/php includes pre-configured PHP extensions and settings for enhanced performance and security. Optimized for Laravel and WordPress.",
  "org.opencontainers.image.documentation": "https://serversideup.net/open-source/docker-php/docs/",
  "org.opencontainers.image.licenses": "GPL-3.0-or-later",
  "org.opencontainers.image.source": "https://github.com/serversideup/docker-php",
  "org.opencontainers.image.title": "serversideup/php ()",
  "org.opencontainers.image.url": "https://serversideup.net/open-source/docker-php/",
  "org.opencontainers.image.vendor": "ServerSideUp",
  "org.opencontainers.image.version": "v3.0.2"
}```

### Current Behavior

Following this guide on running [Laravel Task Scheduler](https://serversideup.net/open-source/docker-php/docs/laravel/laravel-task-scheduler) it says to just call use this command `command: ["php", "/var/www/html/artisan", "schedule:work"]` and it should work. I ran a test command into Docker as just `php artisan` to get alternative output, and this was the result.

<img width="1092" alt="image" src="https://github.com/serversideup/docker-php/assets/352113/e405b44b-fe7e-4634-81c2-39aaaae8c5fe">


### Expected Behavior

When not running the default command, the only output that should be is the command output not output that nginx is booted up and accepting (and receiving ping/status requests internally).

### Steps To Reproduce

1. Use `serversideup/php:8.3.6-fpm-nginx-alpine` as a base
2. Run `docker run ...` and pass a command off to override the built in one per instructions in docs
3. See nginx output

### Anything else?

_No response_

cappuc:
Hi, I've the same problem as describe in #373 and I think I found a good solution to fix this problem:

- use `docker-php-serversideup-entrypoint` as entrypoint like other images instead of `/init` from s6
- don't run the `entrypoint.d` scripts as s6 service because they are run inside `docker-php-serversideup-entrypoint`
- set `/init` as default docker `CMD`

I removed the `docker-php-serversideup-s6-init` since it's not needed anymore


jaydrogers:
Thanks for putting some effort to solve this problem.

I have a few things to update in the documentation (I have learned a lot since I wrote it), but one way of solving this is by setting `--entrypoint=''` on a `docker run` command or `entrypoint: ''` on a Docker Compose service. This works great for things like running a queue that doesn't need FPM or NGINX started.

The `/init` stuff is required by S6 Overlay.

The changes you're proposing might have a big domino effect in other areas if I understand everything correctly.

Do you have a lot of experience with S6 Overlay? I have a bit of experience with it, but there are others who are beyond experts at it 😃

cappuc:
With this changes we don't need to override the entrypoint with `--entrypoint=''` anymore and we can simply override the command like we can do with non fpm images. 
If we override the docker entrypoint, `entrypoint.d` scripts are not executed.

I've already used s6 in other projects but I'm not an expert so maybe I'm missing something.
With the tests I did with this changes, it seems to work by moving `/init` to the command.

Running the image without custom command:
1. execute `entrypoint.d` scripts
2. execute `/init` wich starts nginx + fpm

Running the image with a custom command:
1. execute `entrypoint.d` scripts
2. execute the custom command without starting nginx and fpm

jaydrogers:
That's cool to know. The only thing I am really hesitant on is the docs explicitly say you should set the entrypoint. I've worked on some issues with the founder of S6 and have been called out to "do exactly as I say", which makes sense from an OSS maintainer's perspective. 

I'm just really worried about unintentional side effects if we start using S6 the way it's not intended to be used.

https://github.com/just-containers/s6-overlay?tab=readme-ov-file#usage

Screenshot 2024-08-15 at 08 43 30@2x

Thoughts?

cappuc:
I don't know if there are side effects but I've already used s6 this way in one of my docker images for laravel and it worked. 
In this way we can simply override the command (for example to run horizon or the scheduler) and s6 will not start so nginx and php fpm are not booted.

I found this issue https://github.com/just-containers/s6-overlay/issues/587 on s6 and he suggest to call `/init` conditionally when you want to start s6 services. (I think the code in the last comment is not required because the `CMD` is passed as argument to the entrypoint so we can set `/init` as `CMD` and simply run the argument command)

jaydrogers:
I am just adding a note that I haven't forgotten about this -- I've just been slammed with other things.

I really like where this is going. I am hoping to get a direction on this in the next few days to a week. I will keep you posted 👍

jaydrogers:
I really like what you proposed here! I am moving this into my PR for further testing https://github.com/serversideup/docker-php/pull/437

Thanks a ton for bringing this up. I think it will be a great DX improvement 👍

PR #416, opened 2024-08-21
Add ability to tweak Unit variations max body size setting

RyanTheAllmighty:
Currently the Unit variation doesn't allow customising the `max_body_size` setting (https://unit.nginx.org/configuration/#settings:~:text=false%20(disabled).-,max_body_size,-Maximum%20number%20of), which by default only allows 8MB requests to be processed by Nginx Unit.

This PR adds the ability to specify `UNIT_MAX_BODY_SIZE` environment variable (with a default of 100MB to match PHPs max size).

I've tested this by tweaking the json templates locally and then applying them to my own Docker image and this fixes the issue I was seeing with 413 errors when post sizes were above 8MB.


jaydrogers:
Thanks a ton for submitting this!

This will be added in the v3.4 release 👍

PR #456, opened 2024-10-04
chore: fix typo in docs for new healthcheck script

chaseconey:
## Overview

Quick PR to address an inconsistency in the docs vs what is in the container.

The docs specify the health check for the scheduler as `["CMD", "healthcheck-scheduler"]`, but the actual file name is `healthcheck-schedule`.

### DIFF
--- a/docs/content/docs/3.guide/3.using-healthchecks-with-laravel.md
+++ b/docs/content/docs/3.guide/3.using-healthchecks-with-laravel.md
@@ -87,7 +87,7 @@ label: Using Healthcheck with Laravel Scheduler
 ---
 ```yaml
 healthcheck:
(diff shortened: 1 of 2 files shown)


chaseconey:
Side note, I am not seeing the scripts in the latest version of the container, and they look relatively new. My guess is that they aren't pushed into the stable versions of the image yet, but seeing that in the docs makes it seem like they are. Food for thought.

jaydrogers:
Sorry for the confusion, you're right. The main branch publishes our production website but it also builds `beta` images. I have to re-think that logic.

I appreciate your fix!

Hoping to get these images merged to stable soon so the docs wont be so confusing 😅

PR #462, opened 2024-10-08
Feat: option to specify laravel migration connection

aagjalpankaj:


aagjalpankaj:
In our organization every Laravel application has two DB connections - application, deploy.  application user has DML and deploy user has DDL permissions. Default connection is `DB_CONNECTION=application` set in `.env`. While executing migrate command we can specify which connection we want to use. This PR allows us to set the same as environment variable.

@jaydrogers / @danpastori What do you think about this change?

jaydrogers:
Thanks for proposing this feature!

Unfortunately I had a major refactor on the Laravel Automations script in progress on a different branch, so I was not able to merge this PR.

**Good news is we will support this in v4.0 🥳**

I expanded on your approach and created a `AUTORUN_LARAVEL_MIGRATION_DATABASE` variable. You can specify the exact connection from your database config that you'd like to use (ie. `mariadb`), but I even allow multiple database connections for the 0.1% of people that would ever need to use that 😃 (ie. `mariadb,pgsql`)

Thanks for proposing this feature ❤️

Test it out here:

PR #464, opened 2024-10-13
Replace `exit` with `return` in entrypoint scripts

cappuc:
Calling `exit` in entrypoint scripts prevents subsequent scripts to run.

I noticed the issue withe the `v3.4` beta when running a custom command because `10-init-webserver-config.sh` calls `exit 0` and prevents `50-laravel-automations.sh` to run.


jaydrogers:
Thanks for your help on this! There is a chance I might change `return 1` back to `exit 1` since I want the container to stop and fail at that point.

cappuc:
I thought about it too, but I think it works with `return` too because `docker-php-serversideup-entrypoint` will exit anyway if return code is not 0

Issue #477, opened 2024-10-19
Deployment succeeds when migration fails


aagjalpankaj:
### Steps To Reproduce

1. Set AUTORUN_ENABLED="true".
2. Write a Laravel migration that can fail.
3. Deploy the application. The migration will fail, but the deployment will succeed. 

In such case, code changes will be deployed, but schema changes will not, which may break the application state.

#### What did you expect?
Deployment should be rollbacked or should not succeed when migration fails
#### What happened instead?
Code changes deployed / Deployment succeeded

### Affected Docker Images

I guess most of the images if people are using it for Laravel application with `AUTORUN_ENABLED="true"` option.

jaydrogers:
Just adding a note this will be fixed in v4.0:

- https://github.com/serversideup/docker-php/pull/283

Linking my explanation here: https://github.com/serversideup/docker-php/pull/478#issuecomment-2715105907

PR #478, opened 2024-10-19
Fix: case migration failed

aagjalpankaj:
### Steps To Reproduce
1. Set AUTORUN_ENABLED="true".
2. Write a Laravel migration that can fail.
3. Deploy the application. The migration will fail, but the deployment will succeed.
In such case, code changes will be deployed, but schema changes will not, which may break the application state.

### Outcome
**What did you expect?**
Deployment should be rollbacked or should not succeed when migration fails

**What happened instead?**
Code changes deployed / Deployment succeeded

### Affected Docker Images
I guess most of the images if people are using it for Laravel application with `AUTORUN_ENABLED="true"` option.

---

This PR fixes https://github.com/serversideup/docker-php/issues/477


aagjalpankaj:
@jaydrogers 👀

angristan (approved):

jaydrogers:
I greatly appreciate your efforts to open an issue an propose a solution on this!

In #511, I worked with another developer to completely refactor and improve the Laravel Automations script. While I was working on that branch, I changed some entrypoint execution strategies that allow the Laravel Automations script to fail:

image


You can see in the image I wrote a bad migration from an image built on #511 and the container exits at the bottom because of the failure (which is our expected result).

Although we won't be merging your proposed solution on this issue, we definitely welcome any future contributions. Thanks for helping the community out!

PR #486, opened 2024-10-30
Added pm.max_requests to pool

ifaridjalilov:
You can set with PHP_FPM_PM_MAX_REQUESTS.
Default 1000


angristan (approved):
Thanks! I was searching for this option as well.

Let's keep in mind the default seems to be 0 (https://www.php.net/manual/en/install.fpm.configuration.php):

> pm.max_requests (int)
> The number of requests each child process should execute before respawning. This can be useful to work around memory leaks in 3rd party libraries. For endless request processing specify '0'. Equivalent to PHP_FCGI_MAX_REQUESTS. Default value: 0.

So we might want to keep that.

angristan:
Also `process_idle_timeout` might be cool to have as well.

> pm.process_idle_timeout
> The number of seconds after which an idle process will be killed. Used only when pm is set to ondemand. Available units: s(econds)(default), m(inutes), h(ours), or d(ays). Default value: 10s.

thueske:
Very cool! Looking for exactly that feature. In my opinion, however, the default should also be 0, as this is also in the FPM default.

The documentation should be added here, it's missing in the PR (@ifaridjalilov):
https://github.com/serversideup/docker-php/blob/main/docs/content/docs/7.reference/1.environment-variable-specification.md?plain=1

jaydrogers:
Thanks for raising this feature @ifaridjalilov.

Since #513 has the documentation and the proper defaults, I am going to use that PR.

Thanks for publishing this though. We will always appreciate any future contributions 👍

PR #487, opened 2024-10-31
Adding Cache-Control header to assets and media files

guillaumebriday:
I added the `Cache-Control` header to assets and media files.

Most of applications use the cache-busting pattern so to increase performance we can cache assets basically forever and I added the `immutable` directive.

As explained in the Mozilla documentation:

> When a user reloads the browser, the browser will send conditional requests for validating to the origin server. But it's not necessary to revalidate those kinds of static resources even when a user reloads the browser, because they're never modified. immutable tells a cache that the response is immutable while it's [fresh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#fresh_and_stale_based_on_age) and avoids those kinds of unnecessary conditional requests to the server.

For instance, GitHub or Linkedin are already using it in production.

See: 
- https://www.keycdn.com/blog/cache-control-immutable
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### DIFF
--- a/src/variations/fpm-nginx/etc/nginx/server-opts.d/performance.conf
+++ b/src/variations/fpm-nginx/etc/nginx/server-opts.d/performance.conf
@@ -12,7 +12,7 @@ location = /robots.txt {
 
 # assets, media
 location ~* \.(?:css(\.map)?|js(\.map)?|jpe?g|png|gif|ico|cur|heic|webp|tiff?|mp3|m4a|aac|ogg|midi?|wav|mp4|mov|webm|mpe?g|avi|ogv|flv|wmv)$ {
-    expires    7d;
+    add_header Cache-Control "public, max-age=31536000, immutable";
     access_log off;
     log_not_found off;
     # Pass to PHP to ensure PHP apps can handle routes that end in these filetypes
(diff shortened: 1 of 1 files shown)


jaydrogers:
Today I learned!! Thanks for sharing!

I have a few open source things with [Spin](https://serversideup.net/open-source/spin/) that I'm trying to get out the door, but I will circle back and review once I have a chance. I really like where this is going.

angristan (approved):
I would love this as well 👍 `expires` is the old way to set cache headers: https://stackoverflow.com/questions/5799906/what-s-the-difference-between-expires-and-cache-control-headers

jaydrogers:
I am evaluating this for v3.6 and will keep you posted 👍

jaydrogers:
This all looks great to me! I appreciate you both providing such factual resources for me to learn on this.

I was able to run some tests locally and my "Haha Business" guy had the exact caching experience that I wanted.

I will get this merged in 3.6 👍

Screenshot 2025-03-11 at 12 32 40@2x

guillaumebriday:
Awesome! Thanks for the merge 👍

PR #504, opened 2024-12-18
Referrer policy fix

YevheniiVolosiuk:
I am getting into the console this 

image

to make it better recommends 

image

after adding 

`Header always set Referrer-Policy "strict-origin-when-cross-origin"`

all notices are gone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 # Ensure node files do not appear
+.idea
+vscode
 package-lock.json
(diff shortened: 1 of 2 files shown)


jaydrogers:
Sorry for letting this go stale. Good news is I fixed this in:

- https://github.com/serversideup/docker-php/pull/631

I look forward to seeing another PR from you next time. Thanks again!

PR #511, opened 2025-01-15
Refactor Laravel Autorun Script, adjust entrypoint behavior, contianer info improvements

aSeriousDeveloper:
# What this PR does
This PR started as a simple report by @aSeriousDeveloper, and then the scope was expanded by @jaydrogers 😃

####  Original Comment by @aSeriousDeveloper 

> As discussed in #510 .
> 
> This draft PR will allow for an additional, optional environment variable that will run `php artisan optimize` instead of the usual separate cache commands.
> 
> The purpose of this is that other dependencies hook into this command which cannot be easily accounted for in the autorun. However, the fact that the individual cache functions can be configured to be enabled or disabled means we can't simply switch to `optimize`.
> 
> Instead, we can check if the new environment variable `AUTRORUN_LARAVEL_OPTIMIZE` is set to true. If it is, we simply run `artisan optimize`. If the variable is not set (or is set to false), then call the `cache_laravel` function, which performs the same logic as always.
> 
> By default, `AUTRORUN_LARAVEL_OPTIMIZE` will be resolved to false if not specified, so the original & default configuration for the Dockerfile will be preserved.

As we started working through this process, many other areas were discovered to be improved upon as well.

## 🤩 What's new
- Added `php artisan optimize` support to the Laravel Automations script
- Added `AUTORUN_DEBUG` environment variable to help diagnose issues with Laravel Automations
- Refactored entire Laravel Automations script for better flexbility and user experience
- Changed approach to executing `entrypoint.d` scripts so we can gracefully handle `exit 0` in a entrypoint script
- Re-designed container start up info script

## 🐛 Bug fixes
- Fixed issue where Alpine images were the `/run` directory was not being changed when running `docker-php-serversideup-set-file-permissions`


aSeriousDeveloper:
Huh, found out The optimize function has a new argument that'll be useful here from. [v11.38](https://github.com/laravel/framework/releases/tag/v11.38.0).

[The PR is here](https://github.com/laravel/framework/pull/54070).

You can use `--except` to specify specific cache functions you want to skip when running optimize. You could simply call `optimize` instead of the individual commands, and if one of them is set to false include them in the `except`.

Caveat here is using older versions of Laravel, they wouldn't support this.

aSeriousDeveloper:
I think this might be ready for review / discussion now.
As mentioned in the previous comment, Laravel `v11.38.0` and above support the use of `--except` when optimizing. This lets you decide which optimizations to skip.

I've added this into the script so that the `AUTORUN_LARAVEL_{type}_CACHE` variables can still be respected. However, if you're using a version of Laravel < `11.38.0`, this command will fail and `optimize` without exceptions is ran as a fallback. I've made sure to mention this in the docs.

I will mention, I'm not too well-versed in bash, so someone else may need to review & give this some checking.

jaydrogers:
Thanks! I will take a peek once I get a moment.

The challenging part is we cannot use _Bash_ (/bin/bash), we must use _Shell_ (/bin/sh) 🙃

I'll review and see what we have going on. I want to make sure the user experience is great for existing users (especally ones before Laravel 11) and the upgrade is seamless.

I'll keep you posted!

aSeriousDeveloper:
Made a few changes to make sure it has better compatibility with Shell (it does use POSIX right?), and with the script getting a bit bigger from adding the extra logic I decided to refactor the artisan calls into their own functions. I can reverse the refactoring if preferred.

docs/content/docs/4.laravel/1.laravel-automations.md
```diff
@@ -26,6 +26,20 @@ You can enable the [`--isolated`](https://laravel.com/docs/11.x/migrations#runni
 ### php artisan storage:link
 This command creates a symbolic link from `public/storage` to `storage/app/public`.
 
+### php artisan optimize
+
+This command performs the cache functions mentioned below, as well as provides a way for external dependencies to also cache & optimize their data. This is disabled by default, as Laravel < v11.38.0 does not support granular optimizationvia `--except`.
```
schnetzi, line 31:
```suggestion
This command performs the cache functions mentioned below, as well as provides a way for external dependencies to also cache & optimize their data. This is disabled by default, as Laravel < v11.38.0 does not support granular optimization via `--except`.
```

jaydrogers:
Thanks! Yes, it should be /bin/sh and POSIX compliant.

I'll review sometime (likely next week) as this week has been a busy client week. I greatly appreciate you tackling this!

I also just approved the workflow, so hopefully it will build some preliminary images to test with on DockerHub

src/common/etc/entrypoint.d/50-laravel-automations.sh
```diff
@@ -1,6 +1,22 @@
 #!/bin/sh
 script_name="laravel-automations"
 
+# Set default values for Laravel automations
+: "${AUTORUN_ENABLED:=false}"
+: "${AUTORUN_LARAVEL_STORAGE_LINK:=true}"
+
+# Set default values for optimizations
+: "${AUTRORUN_LARAVEL_OPTIMIZE:=false}"
```
erikgaal, line 9:
```suggestion
: "${AUTORUN_LARAVEL_OPTIMIZE:=false}"
```

src/common/etc/entrypoint.d/50-laravel-automations.sh
```diff
@@ -32,100 +48,136 @@ test_db_connection() {
...
-                echo "✅ Storage already linked..."
-            else
-                echo "🔐 Linking the storage..."
-                php "$APP_BASE_DIR/artisan" storage:link
-            fi
+############################################################################
+# artisan optimize
+############################################################################
+artisan_optimize() {
+    if [ "$AUTRORUN_LARAVEL_OPTIMIZE" = "true" ]; then
```
erikgaal, line 113:
```suggestion
    if [ "$AUTORUN_LARAVEL_OPTIMIZE" = "true" ]; then
```

jaydrogers:
@aSeriousDeveloper: Thank you **SO MUCH** for this PR. It wasn't until I saw your PR is when I realized how UGLY my Laravel automations script has become.

I took what you had and added a lot more flexibility, sanity checks, modularity, etc.

I feel pretty good where this is at so far in my firsts tests, but I have some things to double check before I merge this into 3.6.

Feel free to take a look at what I have going so far and let me know your thoughts.

aSeriousDeveloper:
The automation script really did receieve an overhaul 😸
One thing of note is the docs around `ARTISAN_OPTIMIZE` might need changing a bit, as its condition is now slightly different it seems. I can include that.

erikgaal (approved):
Love it!

PR #517, opened 2025-02-05
feat: run artisan optimize on container start

erikgaal:
Instead of manually running each optimize step, we can leverage the `php artisan optimize` command to run those steps for us.

Also makes use of the new `ServiceProvider::optimizes` API that package developers can use. 😄


erikgaal:
Wow, I'm an idiot and #511 was already created. 🚀

PR #527, opened 2025-04-11
Add default Caddyfile for FrankenPHP that mimics the FPM-NGINX experience

hookenz:
# What this PR does
- It takes the features from our FPM-NGINX confiugrations and replicates this behavior with FrankenPHP

# Related PR
The current PR is for preparation to be merged into:

# Feature status
- [x] Configure SSL with `SSL_MODE`
- [x] Configure default healthchecks
- [x] Ensure `PHP_OPCACHE_ENABLE` works
- [x] Native CloudFlare support with Real IP
- [x] Best practices security headers
- [x] Allow simple troubleshooting with `LOG_OUTPUT_LEVEL`
- [x] Ensure best practice performance/compression is set

### Good examples
- Chris Fidao had a great example: https://gist.github.com/fideloper/af2004b323a4870b2d389e762d631f69

# Original post by @hookenz
This post was edited by @jaydrogers, but the original post by @hookenz is preserved here:

<details>
<summary>@hookenz original Post</summary>

> Hey, I've been using FrankenPHP for a while and it's fantastic.  So much better than Unit.  And just like Unit it's single binary and a one container option.
> 
> This initial Caddyfile sets up the server to listen on 8080 like your other containers.
> It also sets up logging to stdout and creates a healthcheck endpoint with logging turned off for the healthchecks.
> 
> There are some variables, some of which are inherited from frankenphp as I based the caddyfile from theirs.
> 
> `$FRANKENPHP_CONFIG` - optional config for frankenphp.   
> 
>    e.g. from their website:   enable worker mode - ENV FRANKENPHP_CONFIG="worker ./public/index.php"
> 
> $CADDY_GLOBAL_OPTIONS - maybe we should rename this?  It's inherited from the original Caddyfile that comes with the Caddy server.
> 
> What I added: 
> $HEALTHCHECK_PATH - like the other images you can change it away from the default /heathcheck
> $FRANKENPHP_WEBROOT  - defaults to /var/www/html/public/   Just like your other images.  
> 
> Because FRANKENPHP_CONFIG already exists in FRANKENPHP documentation, I kept the same convention of calling it FRANKENPHP rather than FRANKEN_PHP or FRANKEN
> 
> I hope that gets us a bit further down the road.

</details>


hookenz:
@jaydrogers it aint quite right yet.   I'm adding more to this branch.   Caddy syntax is a bit different too.
I think the syntax examples I find on the web might be for caddy v1 sometimes.... so then things don't work.

I'm trying to figure out the best way to enable SSL.  There is also an automatic tls but it's not working for me.
But I can get the non-ssl version working.

hookenz:
Ok, figured out the automatic https.   Now I need to figure out how to best structure the caddyfile

hookenz:
@jaydrogers I think you should be using the franken "builder" images.   Giving it a go on my WIP PR

jaydrogers:
Thanks @hookenz! I greatly appreciate your help to dive in and test some things for me.

Recently I've been buried in client projects, but I will swing back to this as soon as I get a chance. I've had a lot of thoughts on Franken since I originally put my first PR together.

Will keep you posted 👍

sawirricardo:
@hookenz it's been a while and actually frankenphp has some nice updates that have impact on this PR.

hookenz:
I think we should either build the Caddyfile externally or like php unit, start with a simple config file and enable tls config via curl calls to it's api.  I think it can be much simpler than it is.   

To enable automatic tls you don't put the port numbers, just a servername and it's done.

Also the output from caddy can be a bit hard to read.   For an ever so slightly more human readable form set this globally.

```
log {
	format console
}
```

Also discovered that if you pass "-t" to the container, caddy will detect tty use the above automatically.

jaydrogers:
Thanks for waking this PR up. It came at a perfect time.

I have a ton of thoughts, but I need to run out the door right now 🤪

Let me circle back to this, get the merge conflicts fixed, and see if we can get this merged into my new approach to building FrankenPHP

jaydrogers:
Thanks @hookenz for giving me a place to start! I learned a lot with Caddyfile snippets and was able to make a pretty cool process to get this going.

I just merged into the main FrankenPHP branch for further testing 👍

PR #539, opened 2025-06-05
FEAT: add `NGINX_IPV6` env to toggle nginx IPv6 listening

yuuzukatsu:
Hello,

I've been running into some problem with the default configuration listening to IPv6. I was hoping with this env, it enables user to toggle whether they want to listen on IPv6 address or not

image

adding NGINX_IPV6=false environment will delete line containing `listen [::]:` on both `/etc/nginx/site-opts.d/http.conf` and `/etc/nginx/site-opts.d/https.conf`

I've tested this entrypoint script changes specifically in `8.2-fpm-nginx` but adding few lines like this should work in most nginx variation

Thank you

### DIFF
--- a/src/s6/etc/entrypoint.d/10-init-webserver-config.sh
+++ b/src/s6/etc/entrypoint.d/10-init-webserver-config.sh
@@ -55,6 +55,12 @@ process_template() {
     echo "($script_name): Processing $template_file → $output_file..."
     envsubst "$subst_vars" < "$template_file" > "$output_file"
 
+    #Remove line containing listen [::]: to disable IPv6 listening
+    if [ "$NGINX_IPV6" == "false" ]; then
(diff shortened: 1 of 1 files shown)


tomschlick:
It might be a good idea to add `NGINX_IPV4` toggle to this as well. Many K8s environments are IPv6 only (LB / NAT GW handling IPv4 ingress/egress) and could benefit from turning it off.

jaydrogers:
Thank you @yuuzukatsu for the idea on this, and thanks to @tomschlick for chiming in with more use cases.

Instead of `NGINX_IPV6` being a boolean, I created `NGINX_LISTEN_IP_PROTOCOL` (default `all`). You can switch it between `all`, `ipv4`, and `ipv6` to get your desired results.

This will be made available in V4.0:

PR #546, opened 2025-07-03
feat: env configurable self signed cert

joshglasspool:
Makes it possible to configure the auto-generated self-signed cert via env variables if desired. Maintains the existing defaults for the self-signed cert or ability to provide cert and key file instead. Updated docs to detail the new variables.


tomschlick:
IMO the php container shouldn't be responsible for creating a self signed cert as that would only be needed on development. You should use another container to do that and then share the certificate via a mount to the php container.

jaydrogers:
I agree with @tomschlick. I just put default stuff in there just so something would generate.

Any specific use cases on this @joshglasspool?

joshglasspool:
For testing multiple build candidates it’s handy to have a few different self-signed certs. Since the container already makes one, I figured following the convention of making it env configurable was a natural extension. If the consensus is that a sidecar is the right way to go, then it probably makes more sense to just drop cert generation entirely.

jaydrogers:
Thanks for sharing your use case!

I can definitely see this being value in local development environments, especially with `SSL_SUBJECT_CN: "*.dev.test,*.gitpod.io,*.ngrok.io,*.nip.io"`.

If you don't use `*.dev.test` in local environments and you put this behind Traefik, Traefik won't route the traffic becasue of SNI.

Let me stew on this and see if I get this into 4.0 or 4.1. Thanks! 👍

jaydrogers:
I am doing some related work on the v4.0 release and did some testing.

Unfortunately I won't be merging this PR, but the good news is we're going to simplify this even more:

```bash
openssl req -x509 \
    -subj "/CN=localhost" \
    -nodes -newkey rsa:2048 \
    -keyout "$SSL_PRIVATE_KEY_FILE" \
    -out "$SSL_CERTIFICATE_FILE" \
    -days 365 >/dev/null 2>&1
```

I guess through some research, you shouldn't provide multiple values in the Common Name field anyways.

So long story short, we're stripping this out and making it even simpler. I appreciate your efforts to improve the project! I look forward to looking at a PR from you again in the future 👍

PR #547, opened 2025-07-16
introduce start-period and start-interval parameters to HEALTHCHECK

### LINKED ISSUE #541: Increase start_period for default healthcheck
## Background
During zero-downtime deployments, sometime the service fails to start:

```
[09-Jun-2025 17:16:27] NOTICE: fpm is running, pid 131
[09-Jun-2025 17:16:27] NOTICE: ready to handle connections curl: (7) Failed to connect to localhost port 8080 after 1 ms: Could not connect to server ❌ There seems to be a failure in checking the NGINX + PHP-FPM. curl: (7) Failed to connect to localhost port 8080 after 1 ms: Could not connect to server HTTP Status Code: 000 ::1 - -
[09/Jun/2025:17:16:30 +0000] "GET /up HTTP/2.0" 200 1936 "-" "curl/8.12.1" "-" ✅ NGINX + PHP-FPM is running correctly. ::1 - - 
[09/Jun/2025:17:16:33 +0000] "GET /up HTTP/2.0" 200 1936 "-" "curl/8.12.1" "-" ::1 - - 
[09/Jun/2025:17:16:38 +0000] "GET /up HTTP/2.0" 200 1936 "-" "curl/8.12.1" "-"
```

This can mainly be from the Laravel Auto-runs.

From a user on Discord:

> did more digging, finally got a deployment to work, but now I've ran into an old issue that subsequent deployments will always fail and then get rolled back
> 
> the container typically stays alive for about 30 seconds, gets through most of the Laravel automations, and sometimes gives the FPM error I shown above before being reported as a failure and being rolled back
> 
> I am still at a loss as to what's causing this, at first I thought it could be fpm printing to stderr, but that wasn't the case
> 
> it even sometimes starts giving successful health checks before it gets rolled back
> 
> docker inspect implies the container is exiting with code 137 (SIGTERM?)

## Problem

- Our containers are using [the default `start_period`](https://docs.docker.com/reference/dockerfile/#healthcheck) of `0s`, which could create a lot of headaches

## Proposed solution
- Set the `start_period` to `30s`

From the [official docs](https://docs.docker.com/reference/dockerfile/#healthcheck):

> 
start period provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries. However, if a health check succeeds during the start period, the container is considered started and all consecutive failures will be counted towards the maximum number of retries.

#### What did you expect?
- If `AUTORUN_ENABLED` is `true`, we should give enough time for these services to start.

#### What happened instead?
- The container shuts down

### Affected Docker Images

All except CLI

### Related Discord Message
https://discord.com/channels/910287105714954251/910299290230997003/1381686793522380850

aSeriousDeveloper:
- start period allows for a grace period for the container to boot
- start-interval allows for more frequent checking during this grace period
- also slightly increased the standard interval once the container is healthy

Resolves #541 

See [my comment](https://github.com/serversideup/docker-php/issues/541#issuecomment-2960481898) for reasoning behind the values I've chosen.


jaydrogers:
Sorry for the hiccup here, but could you try creating a new PR?

I had to move things in GitHub it automatically closed this PR. I don't see where I can update the branch into https://github.com/serversideup/docker-php/tree/280-create-a-frankenphp-variation

It won't let me change it when I press edit.

jaydrogers:
NVM, I got it. I accidentally deleted the source branch. That's why it closed. I will be reviewing this for 4.0 and will keep you posted 👍

Sorry for the confusion!

jaydrogers:
I love this proposal! Thanks a ton for sharing this, I learned a lot.

I summarized the changes here. Your proposal is brilliant 💯

| **Option** | **Description** | **Old Value** | **New Value** |
| --- | --- | --- | --- |
| start-period | **start period** provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries. However, if a health check succeeds during the start period, the container is considered started and all consecutive failures will be counted towards the maximum number of retries. | - | 60s |
| start-interval | **start interval** is the time between health checks during the start period. | - | 3s |
| timeout | If a single run of the check takes longer than **timeout** seconds then the check is considered to have failed. | 3s | 3s |
| retries | It takes **retries** consecutive failures of the health check for the container to be considered `unhealthy`. | 3 | 3 |
| interval | The health check will first run **interval** seconds after the container is started, and then again **interval** seconds after each previous check completes. | 5s | 10s |

PR #552, opened 2025-08-19
Update GitHub Actions workflow to use depot-ubuntu runners

jaydrogers:
- Changed the runner from ubuntu-24.04 to depot-ubuntu-24.04 for the setup-matrix job.
- Updated the docker-publish job to use depot-ubuntu-24.04-4 as the runner.
- Removed commented-out cache configuration for S3.


jaydrogers:
Closing this and taking a different approach. Might actually move this into stable sooner than later.

PR #558, opened 2025-08-22
feat: allow nginx `client_max_body_size` to be configurable

dlundgren:
- Adds a new `NGINX_CLIENT_MAX_BODY_SIZE` option
- Moves the `client_max_body_size` from the server templates to the nginx.conf template.


jaydrogers:
Sorry for the hiccup here, but could you try creating a new PR?

I had to move things in GitHub it automatically closed this PR. I don't see where I can update the branch into https://github.com/serversideup/docker-php/tree/280-create-a-frankenphp-variation

It won't let me change it when I press edit.

jaydrogers:
NVM, I got it. I accidentally deleted the source branch. That's why it closed. I will be reviewing this for 4.0 and will keep you posted 👍

Sorry for the confusion!

jaydrogers:
This looks great, thank you for your help!

### Quick change I made
- Default is  ~`2G`~ → `100M`

I set it to 100M to match what we have for `PHP_MAX_UPLOAD_SIZE`. I even considered just having `PHP_MAX_UPLOAD_SIZE` in the NGINX configuraiton, but I figured there would be that ONE person who would need to have it separated 🤪

### Why 100M?
This is the default value for the CloudFlare free plan. Also, setting 2G (from what we had before) might have been an old config that I had from years ago. We don't want that number too high so people can blow up a server through a DDoS attack 🤓

Thanks for your contribution!

jaydrogers:
Also, cool to see the contribution come from a fellow Milwaukee developer! 🍻🧀

PR #571, opened 2025-10-10
Fix missing procps on FrankenPHP Variation

arnaud-ritti:
For FrankenPHP variations Laravel healthcheck didn't work decause pgrep was missing.


jaydrogers:
Thanks for pointing this out!

Not sure if this whole PR was AI submitted, but we don't need `gettext` or `zip` for FrankenPHP.

I removed those and will get this published in Beta2 👍

arnaud-ritti:
My bad i had copy paste my lines

PR #602, opened 2025-11-10
Fix "missing unzip" command in FrankenPHP images

### LINKED ISSUE #261: As there is no 'unzip' nor '7z' command installed zip files are being unpacked using the PHP zip extension.
### Affected Images
- Debian images

#### Error Message
Screenshot 2023-12-12 at 10 00 13@2x

```
As there is no 'unzip' nor '7z' command installed zip files are being unpacked using the PHP zip extension.
This may cause invalid reports of corrupted archives. Besides, any UNIX permissions (e.g. executable) defined in the archives will be lost.
```
### Expected Behavior

This error should not appear.

### Steps To Reproduce

Use `spin new laravel demo` to create a new project. This will appear on Composer Install

### Host Operating System

N/A

### Docker Version

```markdown
N/A
```

### Anything else?

_No response_

kohenkatz:
This fixes #261 in the FrankenPHP images. (See also b9118b42b049ac23027785026447396fbcdd0384.)

### DIFF
--- a/docs/content/docs/1.getting-started/7.changelog.md
+++ b/docs/content/docs/1.getting-started/7.changelog.md
@@ -10,4 +10,4 @@ All our changes are documented and published on our GitHub. [See our release not
 ### Choosing a version
 You may want to review [our guide on selecting the right image](/docs/getting-started/choosing-an-image) to determine which version and image tag is best for you.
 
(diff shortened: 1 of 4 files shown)


jaydrogers:
Good catch, thank you 🙏

Issue #607, opened 2025-11-16
fpm-apache variants fail to start on v4.0.0-beta3 when SSL is off


ikerls:
### Steps To Reproduce

 1. Run `docker run --rm -v $(pwd):/var/www/html -p 8080:8080 serversideup/php:8.4.14-fpm-apache-v4.0.0-beta3` into a PHP project

#### What did you expect?
A fully functional PHP server that processes PHP files through PHP-FPM.

#### What happened instead?
Error: 
```
[core:error] [pid 127:tid 127] (EAI 2)Name or service not known: AH00547: Could not resolve host name *:${APACHE_HTTP_PORT:8080} -- ignoring!
```

This makes apache ignores the VirtualHost configuration, resulting in all files (including PHP) being served as static files instead of being processed by PHP-FPM.

### Affected Docker Images

All fpm-apache versions in v4.0.0-beta3

### Anything else?

Full log:
```
--------------------------------------------------------------------
 ____                             ____  _     _        _   _
/ ___|  ___ _ ____   _____ _ __  / ___|(_) __| | ___  | | | |_ __
\___ \ / _ \  __\ \ / / _ \  __| \___ \| |/ _` |/ _ \ | | | |  _ \
 ___) |  __/ |   \ V /  __/ |     ___) | | (_| |  __/ | |_| | |_) |
|____/ \___|_|    \_/ \___|_|    |____/|_|\__,_|\___|  \___/| .__/
                                                            |_|

Brought to you by serversideup.net
--------------------------------------------------------------------

📚 Documentation: https://serversideup.net/php/docs
💬 Get Help: https://serversideup.net/php/community
🙌 Sponsor: https://serversideup.net/sponsor

-------------------------------------
ℹ️ Container Information
-------------------------------------
📦 Versions
• Image:         v4.0.0-beta3-19117516691
• PHP:           8.4.14
• OS:            Debian GNU/Linux 13 (trixie)

👤 Container User
• User:          www-data
• UID:           33
• GID:           33

⚡ Performance
• OPcache:       ❌ Disabled
• Memory Limit:  256M
• Upload Limit:  100M

🔄 Runtime
• Docker CMD:     /init

👉 [NOTICE]: Improve PHP performance by setting PHP_OPCACHE_ENABLE=1 (recommended for production).
ℹ️ NOTICE (generate-ssl): SSL mode is off, so we won't generate a self-signed SSL key pair.
[...]

PR #608, opened 2025-11-16
fix(config): remove default port from apache http virtualhost

ikerls:
Removes the default value for `APACHE_HTTP_PORT` in `ssl-off.conf` (to maintain consistency with `ssl-full.conf` and `ssl-mixed.conf`) as it is causing the error:
```
Name or service not known: AH00547: Could not resolve host name *:${APACHE_HTTP_PORT:8080} -- ignoring!
```
The default value for `APACHE_HTTP_PORT` is being set in `src/variations/fpm-apache/Dockerfile`.
Fixes: #607


jaydrogers:
Oh dang, I just saw this...

I pushed a commit to fix it, but I just squashed and merged this to see if you can still get the credit.

Thanks for your contribution!

PR #638, opened 2026-01-10
Update PHP base opertating systems. Add Alpine 3.23 support

alloylab:
PHP repo no longer has images available for alpine3.21 for latest minor versions, this removes 3.21 if its no longer available and adds 3.23 if available

### DIFF
--- a/scripts/conf/php-versions-base-config.yml
+++ b/scripts/conf/php-versions-base-config.yml
@@ -35,42 +35,43 @@ php_versions:
       - minor: "8.1"
         base_os:
           - name: alpine3.21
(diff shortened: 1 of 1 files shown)


jaydrogers:
Thank you for your effort on this! It's been a little crazy around here after being on holiday then coming back to a chaotic work schedule 🤪

This was on my list to do, but you nailed it. Thanks for your help!

alloylab:
@jaydrogers Anytime... I understand how it goes.. thanks for merging... i am monitoring the build right now to make sure I didn't missing anything

PR #643, opened 2026-01-22
Add TRUSTED_PROXY support for all variations & Caddy global imports

jaydrogers:
# How to test this image
Use the `serversideup/php-dev` images that are tagged with this PR number:

```bash
serversideup/php-dev:643-*
```

[View the images →](https://hub.docker.com/r/serversideup/php-dev/tags?name=643-)

# What this PR does

### Add `TRUSTED_PROXY` support
You can now customize your trusted proxy experience via environment variables.

| Variation | Affected |
|--|--|
| cli | ❌ |
| fpm | ❌ |
| fpm-nginx | ✅ |
| fpm-apache | ✅ |
| frankenphp | ✅ |

| Value | Description |
|-------|-------------|
| `cloudflare` (default) | Trusts [Cloudflare's IP ranges](https://www.cloudflare.com/ips/) and uses the `CF-Connecting-IP` header |
| `sucuri` | Trusts [Sucuri's IP ranges](https://docs.sucuri.net/website-firewall/sucuri-firewall-troubleshooting-guide/) and uses the `X-Forwarded-For` header |
| `local` | Trusts only private/local network ranges (Docker networks, localhost) and uses the `X-Forwarded-For` header |
| `off` | Disables trusted proxy configuration entirely |

### Global Imports for FrankenPHP
Instead of using `${CADDY_GLOBAL_OPTIONS}` variable (which can be a pain with escaping characters, etc) you can now just import configurations into the global block by copying Caddyfiles over to `/etc/frankenphp/caddyfile-global.d/`.

Caddy will load any `*.caddyfile` files:

https://github.com/serversideup/docker-php/blob/5d5d7a92b6d791723f75f5c7c59b744060b4f737/src/variations/frankenphp/etc/frankenphp/Caddyfile#L22-L23


QarthO:
One recommendation would be for local to be somehow tied into cloudflare/sucuri. With this new implementation, if I have my site behind cloudflare, but also running in docker (behind traefik/caddy), then i shouldnt have to pick/choose if i want both

Not sure what the best way to do this, but i do feel like a docker setup is very common

jaydrogers:
Valid point!

If you check the file, you'll see that when you set `sucuri` it will do the local addresses as well as sucuri.

https://github.com/serversideup/docker-php/blob/f696ba5323e73791b47f6d52c9a69e82ac762719/src/variations/fpm-nginx/etc/nginx/trusted-proxy/sucuri.conf#L1-L22

I just added some docs to make it clearer 😃

## Next steps
If you want to test out using the images in the top comment, let me know if you're getting your expected results 👍

mbrodala:
A bit late but would it make sense to prefer [`RemoteIPInternalProxy`](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy) for the "Docker networks"? Or should this be handled individually per app? (Which then likely requires `TRUSTED_PROXY=off` AFAIS)

Mostly asking because `REMOTE_ADDR` is left unchanged by mod_remoteip in case it contains a private network IP address and only `RemoteIPTrustedProxy` is configured:

> Unlike the [RemoteIPInternalProxy](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipinternalproxy) directive, any intranet or private IP address reported by such proxies, including the 10/8, 172.16/12, 192.168/16, 169.254/16 and 127/8 blocks (or outside of the IPv6 public 2000::/3 block) are not trusted as the useragent IP, and are left in the [RemoteIPHeader](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipheader) header's value.

jaydrogers:
Thanks for pointing this out! I wasn't aware of `RemoteIPInternalProxy`.

I had Fable 5.1 help me on some of this and it raised https://github.com/serversideup/docker-php/issues/703

I applied your fixes and some other fixes to harden the security here: https://github.com/serversideup/docker-php/commit/59b05055ad391730c9cb8c6ba0827257a8ebc2ed

You can test it here if you'd like:

- https://github.com/serversideup/docker-php/pull/645

mbrodala:
Thanks a lot! This fixes the issue. Now given a setup with CF in front and various internal proxies (don't ask) we get the actual client IP address in `REMOTE_ADDR` just fine. :thumbsup:

PR #644, opened 2026-01-26
Do not generate SSL if DISABLE_DEFAULT_CONFIG is set

5ergiu:
Add checks for DISABLE_DEFAULT_CONFIG and set default SSL file paths.


jaydrogers:
Thanks for this, great catch!

This will be made available in the next release 👍

PR #645, opened 2026-01-27
Release 5.0: Add IP address SSL certificates, Additonal Proxy Support, web server improvements, and other fixes

jaydrogers:
## 👨‍🔬 How to test
You can test this image using our serversideup/php-dev repository, which automatically builds on push to this PR.

```
serversideup/php-dev:645-*
```

[View the available testing images →](https://hub.docker.com/r/serversideup/php-dev/tags?name=645-)

## 🌎 Docs Preview
You can view the latest docs for this release here:

- https://release-webserver-improvemen.serversideup-php.pages.dev/open-source/docker-php/

---
## 🤩 What's new

image

### Native Laravel Octane support with a hardened and optimized Caddyfile (#604)
Add `--caddyfile=/etc/frankenphp/Caddyfile` to `octane:start` and FrankenPHP switches into worker mode with the same trusted proxies, security headers, asset caching, SSL modes, and health check as classic mode. Octane's `--workers`, `--watch`, `--max-requests`, `octane:status`, and `octane:reload` all work. No wrapper script, no second Caddyfile.

```yml
command: ["php", "artisan", "octane:start", "--server=frankenphp", "--port=8080", "--caddyfile=/etc/frankenphp/Caddyfile"]
```

Thanks to @aSeriousDeveloper for opening the original PR and coolio85 on Discord for the workaround that started it.

[Read the Octane guide →](https://release-webserver-improvemen.serversideup-php.pages.dev/open-source/docker-php/docs/framework-guides/laravel/octane)

### Expanded OPcache support and better defaults (#704)
OPcache is still off by default to save your hair when you're in development. When you set `PHP_OPCACHE_ENABLE=1`, you now get the settings from [Symfony's performance guide](https://symfony.com/doc/current/performance.html#configure-opcache-for-maximum-performance) instead of PHP's stock values: more memory, room for more files, and no filesystem checks per request. Preloading is available with `PHP_OPCACHE_PRELOAD`. This is a breaking change if you mount your code as a volume with OPcache enabled (see below).

[Read the production performance tuning guide →](https://release-webserver-improvemen.serversideup-php.pages.dev/open-source/docker-php/docs/guide/production-performance-tuning)

### Short-lived & IP address certificates for FrankenPHP
[...]

PR #667, opened 2026-03-29
Add configurable real IP header variables for Apache and NGINX

ps-20x:
### Problem

The real client IP header is currently hardcoded to `CF-Connecting-IP` in both the Apache and NGINX configurations. This works well for deployments sitting directly behind Cloudflare, but users behind other reverse proxies (Traefik, AWS ALB, Kubernetes ingress, generic NGINX, etc.) have to either set `DISABLE_DEFAULT_CONFIG`, mount override files, or patch the shipped config in a derived image — just to change a single directive.

This is especially inconvenient because the rest of the default configuration is usually still desirable; only the header name needs to change.

**Real-world example:** I run a WordPress site on Hetzner with `fpm-apache` behind Traefik as a reverse proxy. Traefik forwards the client IP via `X-Forwarded-For`, but Apache never picks it up because the VHost templates hardcode `CF-Connecting-IP`. The result: WordPress shows the internal Docker network IP (e.g. `172.18.0.5`) in password reset emails and WooCommerce order logs instead of the actual visitor IP. The only workaround was patching the templates in a derived image.

### Changes

**Apache** — new environment variable `APACHE_REMOTE_IP_HEADER` (default: `CF-Connecting-IP`)

Replaces the hardcoded header name in:
- `conf-available/remoteip.conf`
- `vhost-templates/http.conf`
- `vhost-templates/https.conf`
- `sites-available/ssl-full.conf`

Uses Apache's native `${VAR}` substitution at daemon startup — no init script changes needed.

**NGINX** — new environment variable `NGINX_REAL_IP_HEADER` (default: `CF-Connecting-IP`)

Converts the NGINX real IP config into a template:
- `server-opts.d/remoteip.conf`

Processed via the existing `process_template` / `envsubst` mechanism during container init.

**Documentation** — both variables are added to the environment variable specification reference page.

### Backward Compatibility

Both variables default to `CF-Connecting-IP`. Existing deployments are completely unaffected — behavior only changes when a user explicitly sets a different value.

### Usage

```yaml
environment:
  APACHE_REMOTE_IP_HEADER: "X-Forwarded-For"
  # or for NGINX:
  NGINX_REAL_IP_HEADER: "X-Forwarded-For"
```

### Note on Trusted Proxies

This PR only makes the **header name** configurable. The trusted proxy CIDR list remains unchanged. Users switching away from Cloudflare should review whether the trusted proxy configuration matches their infrastructure.

### Scope

This PR is intentionally narrow: configurable header name, preserved defaults, no changes to trusted proxy semantics or existing user behavior.


jaydrogers:
Thanks for your PR! Unfortunately I won't be able to merge this because I had pending changes for the 4.4 release: https://github.com/serversideup/docker-php/pull/645

4.4 incudes these updates: https://github.com/serversideup/docker-php/pull/643

Could you review those changes and see if that answers your questions/goals?

PR #692, opened 2026-08-20
Add PHP_DISABLE_FUNCTIONS, PHP_HTML_ERRORS, PHP_REALPATH_CACHE_SIZE, and PHP_SESSION_COOKIE_HTTPONLY environment variables

Abdulmajeed-Jamaan:
> [!NOTE]
> **Note from @jaydrogers:** I took over this PR to finish it and rewrote this description to match our PR format. See [my comment below](https://github.com/serversideup/docker-php/pull/692#issuecomment-XXXX) for what changed and why. @Abdulmajeed-Jamaan's original description is preserved at the bottom of this comment.

### Why we created this PR

Four common `php.ini` settings could only be changed by mounting a custom ini file. Every other PHP setting in our images is an environment variable, so these should be too. @Abdulmajeed-Jamaan opened this PR so they could delete their custom ini file, and we agreed.

### How to test

```
serversideup/php-dev:692-*
```

[View the testing images →](https://hub.docker.com/r/serversideup/php-dev/tags?name=692-)

Confirm a variable reaches PHP:

```sh
docker run --rm -e PHP_REALPATH_CACHE_SIZE=8M serversideup/php-dev:692-8.4-cli \
  php -r 'echo ini_get("realpath_cache_size");'
```

### What this PR does

**New environment variables**

| Variable | Default | What it does |
|---|---|---|
| `PHP_DISABLE_FUNCTIONS` | `""` | Comma-separated list of PHP functions to disable, such as `exec,shell_exec`. Empty because Laravel, Composer, and Symfony Process rely on `proc_open`. |
| `PHP_HTML_ERRORS` | `On` | Format on-screen errors as HTML when `PHP_DISPLAY_ERRORS` is on. Never affects logged errors. PHP forces it off for the CLI. |
| `PHP_REALPATH_CACHE_SIZE` | `4096K` | Size of PHP's realpath cache. Apps with a large `vendor/` directory may benefit from raising it. |
| `PHP_SESSION_COOKIE_HTTPONLY` | `On` | Adds the `HttpOnly` flag to the native PHP session cookie so browser scripts cannot read it. |

The first three defaults match what the images already did, so nothing changes unless you set them. `PHP_SESSION_COOKIE_HTTPONLY` is the exception, see Compatibility below.

**Docs**
- Added all four variables to the [environment variable specification](https://release-webserver-improvemen.serversideup-php.pages.dev/open-source/docker-php/docs/reference/environment-variable-specification)
- Fixed the listed default for `PHP_SESSION_COOKIE_SECURE`. The docs said `true`, but the images have always shipped it as `false`.

**Tests**
- `scripts/test-image.sh` now overrides a size, a boolean, and a list variable and checks that PHP reports the new values. Every image in CI proves the ini substitution works.

### Compatibility

`PHP_SESSION_COOKIE_HTTPONLY` now defaults to `On`. PHP [recommends this for production](https://www.php.net/manual/en/session.security.ini.php), and the next PHP release ships it on in `php.ini-production`. Laravel, Symfony, and WordPress manage their own session cookies and are not affected. If your app calls `session_start()` directly and reads the session cookie from JavaScript, set `PHP_SESSION_COOKIE_HTTPONLY=Off` to keep the old behavior.

---

<details>
<summary>Original description by @Abdulmajeed-Jamaan</summary>

Set the the default values from php docs as needed.

This will make me remove the custom php ini file created to configure them, therefore enhancing DX

</details>


bretto36 (approved):

jaydrogers:
Thanks @Abdulmajeed-Jamaan! This was a good idea and I'm glad to have these settings as environment variables.

I took the PR over to get it across the finish line and made a few changes along the way. Here's what and why:

### Added the missing `PHP_DISABLE_FUNCTIONS` env var
The ini file referenced it, but no Dockerfile declared it. It worked only because PHP treats an unset variable as empty, so it never showed up in `docker inspect` or the docs.

### Changed a few defaults
`PHP_REALPATH_CACHE_SIZE` is now `4096K` instead of `4M`. Same value, but it matches PHP's own default and the comment in the ini file. `PHP_HTML_ERRORS` is `On` instead of `1` to match how we write the other boolean settings.

### Turned `PHP_SESSION_COOKIE_HTTPONLY` on by default
PHP recommends it for production and the next PHP release turns it on in `php.ini-production`, so we decided to ship it that way now. I've called this out in the release notes.

### Added docs and a test
All four variables are in the environment variable spec, and `scripts/test-image.sh` now checks that env vars actually reach `php.ini`. I also fixed the alphabetical order in the Dockerfile `ENV` blocks.

Thanks again for the contribution!

PR #697, opened 2026-09-07
Add PHP_FILE_UPLOADS and PHP_MAX_FILE_UPLOADS to environment variables

agabi10:
In most of the containers I have I don't use any of the upload features, and in the ones I do I allow the upload of one file per request at most, so I'd like to be able to restrict the uploads to narrow the vector of attack from my services. This would be way more convenient with environmental variables.


jaydrogers:
This looks great! Thank you for your contribution!

This will be made available in our next release:

- https://github.com/serversideup/docker-php/pull/645

PR #702, opened 2026-09-10
Fix FRANKENPHP_CONFIG environment variable typo in FrankenPHP Dockerfile

Abdulmajeed-Jamaan:
## What this does

Renames the `FRANKEN_PHP_CONFIG` ENV in `src/variations/frankenphp/Dockerfile` to `FRANKENPHP_CONFIG`.

## Why

The Dockerfile declares the default as `FRANKEN_PHP_CONFIG=""`, but everything else in the project uses `FRANKENPHP_CONFIG`:

- `src/variations/frankenphp/etc/frankenphp/Caddyfile` reads `{$FRANKENPHP_CONFIG}` inside the `frankenphp { }` global block
- The [FrankenPHP docs](https://serversideup.net/open-source/docker-php/docs/image-variations/frankenphp) and the [environment variable specification](https://serversideup.net/open-source/docker-php/docs/reference/environment-variable-specification) both document `FRANKENPHP_CONFIG`

The typo was introduced in #527 when the default Caddyfile was added. Setting `FRANKENPHP_CONFIG` at runtime already works because Caddy reads it directly from the environment, so this only corrects the declared default in the image and removes a misleading, unused variable from `docker inspect` output.

## Scope

One-line change. No behavior change for users who already set `FRANKENPHP_CONFIG` themselves.


jaydrogers:
Good catch, thanks!

